Issue #19, opened 2017-07-11
Global exception handler sends along the exception message by default


pedelman:
For global exception handler, the default behavior is to send along the e.message. This behavior was added in PR #10.

See: https://github.com/bigcommerce/gruf/blob/master/lib/gruf/service.rb#L231

Exceptions can contain potentially sensitive server information and we may not want this data to be sent along to clients. We should set up a config to allow for this behavior to more tightly controlled by the service.

Perhaps something like the following.

```rb
Gruf.configure do |c|
  c.internal_error_message = 'Server Error'
  # other config setting
end
```

Either that or we can create a default value and explicitly require the configuration to enable the `e.message` logging. I personally see the `e.message` error as an insecure default.

```rb
Gruf.configure do |c|
  c.internal_exception_message = true # default is false
  # other config setting
end
```

We could also go with a combination of the two options. Any other thoughts?

pedelman:
You can now get around this issue by setting the following in your Gruf config.

```rb

Gruf.configure do |c|
  c.use_exception_message = false
  c.internal_error_message = "Your error message here" # defaults to "Internal Server Error"
end
```

PR #20, opened 2017-07-11
Update configuration and global exception handler to not pass e.message.

pedelman:
## What? Why?

Sending along an uncaught exception message by default can be considered insecure. This can leak server information to clients of the service unintentionally. 

This PR will set the error_message for an uncaught exception and allow a user to both modify the error message as well as enable sending the exception message if that is intended behavior.

## How was it tested?

Rspec tests have been added.

### DIFF
--- a/lib/gruf/configuration.rb
+++ b/lib/gruf/configuration.rb
@@ -38,7 +38,9 @@ module Gruf
       authorization_metadata_key: 'authorization',
       append_server_errors_to_trailing_metadata: true,
       use_default_hooks: true,
-      backtrace_on_error: false
+      backtrace_on_error: false,
+      use_exception_message: true,
+      internal_error_message: 'Internal Server Error'
     }.freeze
 
     attr_accessor *VALID_CONFIG_KEYS.keys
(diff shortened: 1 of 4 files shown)


pedelman:
Fixes #19

lib/gruf/configuration.rb
```diff
@@ -38,7 +38,9 @@ module Configuration
       authorization_metadata_key: 'authorization',
       append_server_errors_to_trailing_metadata: true,
       use_default_hooks: true,
-      backtrace_on_error: false
+      backtrace_on_error: false,
+      use_exception_message: false,
```
splittingred: We should default this to `true`

lib/gruf/configuration.rb
```diff
@@ -38,7 +38,9 @@ module Configuration
...
+      internal_error_message: 'Server Error'
```
splittingred: Can you make this `Internal Server Error`

pedelman:
@splittingred I have addressed your comments

PR #21, opened 2017-07-14
Refactor instrumentation to outer_around chain, add request logging

splittingred:
## What? Why?

This PR adds the following functionality to gruf:

* Instrumentation hooks are now executed in the context of the `outer_around` chain - this allows them to properly pick up failed or unauthenticated requests (as opposed to just "successful" ones, which they did prior
* Instrumentation hooks now pass a `RequestContext` object that contains information about the incoming request, instead of relying on instance variables 
* The StatsD hook now sends metrics for failed and successful, as well as aggregate, calls
* There is now a default RequestLogging hook enabled, which adds Rails-style request logging for gRPC service calls. It has multiple formatters available - currently `plain` (default) and `logstash`. The formatters are built for easy expansion and customization should more need to be added
* Reduced reliance on `to_sym` where possible

## How was it tested?

Through spec suite, and also through multiple tests on various systems and through the `spec/demo_server` client.

----

@bigcommerce/services @pedelman @junedkazi


lib/gruf/instrumentation/request_logging/formatters/base.rb
```diff
@@ -0,0 +1,38 @@
...
+        # Base class for request log formatting
+        #
+        class Base
+          ##
+          # Format the parameters into a loggable string. Must be implemented in every derivative class
+          #
+          # @param [Hash] _payload The incoming request payload
+          # @return [String] The formatted string
+          #
+          def format(_payload)
```
ncancelliere: What's the point of the _ here?  I know that sometimes people use _ prefix to indicate that the variable is unused. Why not just do. `def format(payload={})`. that'd communicate better what it is (unless there's a reason why you don't want to default it to an empty hash).

lib/gruf/instrumentation/request_logging/hook.rb
```diff
@@ -0,0 +1,116 @@
...
+
+        private
+
+        ##
+        # Determine the appropriate formatter for the request logging
+        #
+        # @return [Gruf::Instrumentation::RequestLogging::Formatters::Base]
+        #
+        def formatter
+          fmt = options.fetch(:formatter, :plain)
```
ncancelliere: Is there a reason why we return a new formatter every time?  Why not memoize this?

lib/gruf/instrumentation/statsd.rb
```diff
@@ -26,6 +26,8 @@ class Statsd < Gruf::Instrumentation::Base
       def call
         if client
           client.increment(route_key)
+          postfix = success? ? 'success' : 'failure'
```
ncancelliere: I'd make this a private method called `postfix`.

lib/gruf/instrumentation/request_logging/formatters/base.rb
splittingred: `_` is used here to show it's not used in this class, but still required to be passed. By setting `={}` that would mean it's an optional parameter, which it's not.

lib/gruf/hooks/base.rb
```diff
@@ -45,6 +45,22 @@ def initialize(service, options = {})
...
+      def service_key
+        service.class.name.underscore.tr('/', '.')
+      end
+
+      ##
+      # Parse the method signature into a service.method name format
+      #
+      # @return [String] The parsed service method name
+      #
+      def method_key(call_signature)
```
pedelman: It seems redundant to log the service_key an extra time here.

Here is an example log from a health check service I have running.

```json
{
  "params":{},
  "message":"",
  "service":"health.health_check_service",
  "action":"health.health_check_service.check",
  "method":"health.health_check_service.check",
  "grpc_status":"GRPC::Ok",
  "duration":3.5,
  "status":0,
  "thread_id":46958605007800,
  "time":"2017-08-01 14:27:48 +0000",
  "host":"ebfc65cb3169",
  "format":"json",
  "@timestamp":"2017-08-01T14:27:48.405+00:00",
  "@version":"1",
  "severity":"INFO"
}
```

I think action and method should both just be the gRPC method name without being prefixed with the service_key.

Thoughts?

lib/gruf/instrumentation/request_logging/hook.rb
```diff
@@ -0,0 +1,116 @@
...
+          else
+            type = :error
+            status_name = response.class.name
+          end
+
+          payload = {}
+          payload[:params] = sanitize(request.to_h) if options.fetch(:log_parameters, false)
+          payload[:message] = success? ? '' : response.message
+          payload[:service] = service_key
+          payload[:method] = payload[:action] = route_key
```
pedelman: Do we need to do both a `method` and `action` here?

splittingred: A lot of common logstash filters base on `action`; while others use `method`. Wanted to keep that there to not break backwards compatibility with those filters.

lib/gruf/hooks/base.rb
pedelman: Also can we have a better default message? Here is the log I had previously been writing in a custom hook.

```
{
  "params":{},
  "message":"[GRPC::Ok] (health.health_check_service.check)",
  "service":"health.health_check_service",
  "call":"check",
  "grpc_status":"GRPC::Ok",
  "@timestamp":"2017-08-01T15:11:28.432+00:00",
  "@version":"1",
  "severity":"INFO",
  "host":"ffe831f5cb44",
  "thread_id":47383804903340
}
```

UPGRADING.md
```diff
@@ -0,0 +1,31 @@
...
+Gruf.configure do |c|
+  c.instrumentation_options[:request_logging] = {
+    formatter: :logstash,
+    log_parameters: false
+  }
+end
+```
+
+If you want to log parameters, we recommend setting a blacklist to ensure you don't accidentally
+log sensitive data.
```
pedelman: Might also leave a note about blacklisted params which may contain very large values (such as json strings or binary values)

splittingred:
@ncancelliere @pedelman ♻️ 🙇

PR #24, opened 2017-08-15
Allow GRPC::RpcServer configuration

kruczjak:
## What? Why?
Ability to configure `GRPC::RpcServer` (e.g. `pool_size`) was missing (http://www.rubydoc.info/gems/grpc/GRPC/RpcServer#initialize-instance_method). 

## How was it tested?
Specs and running `bundle exec gruf` with and without additional config.


splittingred:
This looks good! Thanks for your contribution. :+1:

PR #25, opened 2017-09-24
Prevent Gruf::Server from loading non-service files

billthompson:
## What? Why?
Gruf explicitly requires files in the `servers_path` directory. This doesn't play nicely with autoloading in Rails projects when concerns live in the `servers_path` directory as they can only be required once.

This PR:
* Extracts the services loading functionality into `Gruf::ServiceLoader`
* Updates the loading logic to identify  and only require services that include the Gruf::Service mixin

## How was it tested?
`rspec`. Note: I copied the existing service loading specs from `server_spec.rb` to help make sure no regressions are introduced in the refactor.

### DIFF
--- a/lib/gruf/server.rb
+++ b/lib/gruf/server.rb
@@ -22,10 +22,9 @@ module Gruf
   class Server
     include Gruf::Loggable
 
-    # @return [Array<Class>] The services this server is handling
-    attr_accessor :services
+    delegate :services, to: :service_loader
 
-    attr_reader :server
+    attr_reader :server, :service_loader
 
     ##
     # Initialize the server and load and setup the services
@@ -33,8 +32,8 @@ module Gruf
     # @param [Array<Class>] services The services that this server should handle
     #
     def initialize(services: [])
+      @service_loader = ServiceLoader.new(services: services)
       setup!
-      load_services(services)
     end
 
     ##
@@ -65,28 +64,11 @@ module Gruf
     private
 
     ##
-    # Return all loaded gRPC services
-    #
-    # @param [Array<Class>] svcs An array of service classes that will be handled by this server
-    # @return [Array<Class>] The given services that were added
-    #
-    def load_services(svcs)
-      unless @services
-        @services = Gruf.services.concat(svcs)
-        @services.uniq!
-      end
-      @services
-    end
-
-    ##
     # Auto-load all gRPC handlers
     #
     # :nocov:
     def setup!
-      Dir["#{Gruf.servers_path}/**/*.rb"].each do |f|
-        logger.info "- Loading gRPC service file: #{f}"
-        require f
-      end
+      service_loader.require_services
     end
     # :nocov:
 
--- a/lib/gruf/service_loader.rb
+++ b/lib/gruf/service_loader.rb
@@ -1,0 +1,65 @@
+# coding: utf-8
+# Copyright (c) 2017-present, BigCommerce Pty. Ltd. All rights reserved
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+# documentation files (the "Software"), to deal in the Software without restriction, including without limitation the
+# rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit
+# persons to whom the Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions of the
+# Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+# WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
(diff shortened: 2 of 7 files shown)


lib/gruf/service_loader.rb
```diff
@@ -0,0 +1,65 @@
...
+module Gruf
+  ##
+  # Register and require Gruf services so they can be served.
+  #
+  class ServiceLoader
+    include Gruf::Loggable
+
+    SERVICE_IDENTIFIER = 'include Gruf::Service'.freeze
+
+    attr_accessor :services, :load_path
```
splittingred: IMO no need for these variables to be public; let's keep them private

lib/gruf/service_loader.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    private
+
+    ##
+    # Register services with the loader
+    #
+    # :nocov:
+    def register_services(svcs)
+      self.services = Array(svcs).concat(Gruf.services).compact.uniq
```
splittingred: Law of demeter violation here. I'd expect the constructor like this:

```
registry = Gruf::ServiceRegistry.new(services: Gruf.services, load_path: Gruf.servers_path)
registry.load
```

IMO `ServiceRegistry` should have no knowledge of the `Gruf` configuration values and be fully injectable.

lib/gruf/service_loader.rb
```diff
@@ -0,0 +1,65 @@
...
+      @load_path = load_path
+      register_services(services)
+    end
+
+    ##
+    # Require all Gruf::Service's in the load_path. Files in the load_path that aren't
+    # identified as Gruf::Service's will not be required explicitly.
+    #
+    def require_services
+      files_in_load_path = Dir["#{load_path}/**/*.rb"]
```
splittingred: Should return early if `load_path` is empty or directory is non-existent

lib/gruf/service_loader.rb
```diff
@@ -0,0 +1,65 @@
...
+      files_in_load_path.each do |f|
+        file_text = File.read(f)
```
splittingred: `File.read(File.realpath(f))`

lib/gruf/server.rb
```diff
@@ -22,19 +22,18 @@ module Gruf
   class Server
     include Gruf::Loggable
 
-    # @return [Array<Class>] The services this server is handling
-    attr_accessor :services
+    delegate :services, to: :service_loader
 
-    attr_reader :server
+    attr_reader :server, :service_loader
```
splittingred: Let's make `service_loader` private and non-exposed. You can remove the `delegate` call to `services` above as well. I'd rather us not expose that API publicly and provide method contracts _later_ if we feel the need is there - mainly because we may want to implement dynamic server reflection based on remote calls later, and I'd rather that not be writable.

lib/gruf/service_loader.rb
```diff
@@ -0,0 +1,65 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+# WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+# OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+#
+module Gruf
+  ##
+  # Register and require Gruf services so they can be served.
+  #
+  class ServiceLoader
```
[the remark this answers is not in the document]
billthompson: Sure thing; that's what I went with originally but changed to loader to match the naming of the existing functionality. I'll get this refactored.

lib/gruf/service_loader.rb
billthompson: I have made `load_path` a required parameter during ServiceRegistry instantiation and the Dir glob handles paths that don't exist for us automatically:
```
[1] pry(main)> Dir["fake/path/that/does/not/exist/**/*.rb"]
=> []
```

Would you still prefer to see an explicit `return if files_in_load_path.empty?` or are you okay with the iterator handling an empty list?

splittingred: Good point - I think it's fine letting `Enumerable` handle the work here. Thanks!

lib/gruf/service_loader.rb
billthompson: I don't think this is a Law of Demeter violation because `concat`, `compact`, and `uniq` are all being called on the Array - this is just method chaining. If on the other hand I tried to access methods on services inside the array, then that code would be in violation.

But I have refactored the logic that combins the services specified at server instantiation with the services declared during initialization to be performed inside the server - that way the registry doesn't have to be aware of `Gruf.services`.

spec/demo_app_root/non_sevice.rb
```diff
@@ -0,0 +1,3 @@
+module NonService
```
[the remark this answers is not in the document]
billthompson: 😊

splittingred:
Fixed by #28. Closing.

billthompson:
Thanks @splittingred!

PR #27, opened 2017-09-26
#26: Fix issue with field errors being persisted between calls

### LINKED ISSUE #26: Field errors are persisted in between calls
Have been able to confirm this by creating an RPC method that adds field errors depending on incoming value:

```ruby
context 'with multiple calls, where the first returns an error, the second doesnt' do
  it 'should not raise a Gruf::Client::Error the second time', run_thing_server: true do
    client = build_client
    expect do
      client.call(:GetContextualFieldErrorFail, id: 1)
    end.to raise_error(Gruf::Client::Error)

    expect do
      client.call(:GetContextualFieldErrorFail, id: 2)
    end.to_not raise_error(Gruf::Client::Error)
  end
end
```

It will throw an error currently:

```bash
expected no Gruf::Client::Error, got #<Gruf::Client::Error: Gruf::Client::Error> with backtrace:
```

This seems to be due to `@error` being set memoized on the service class, and not reset in-between calls. This was designed in order to support the `add_field_error` method without having to pass around an error object. A simple fix is just to unset the `@error` attribute at the end of every call.

splittingred:
## What? Why?

Fixes #26. Also:

* Fixes the test suite so that the server functional specs can run for each request type
* Adds more functional tests around error handling

## How was it tested?

`rspec`


spec/gruf/error_spec.rb
```diff
@@ -125,4 +125,77 @@
       expect { subject }.to raise_error(GRPC::NotFound)
     end
   end
+
+  describe 'functional test' do
+    let(:id) { 1 }
+
```
billthompson: What do you think about building your client in a `let` var instead of inside each example?

splittingred: It needs to be built in the call itself due to the port being dynamic, and the port is generated by the `run_thing_server` hook each time. Using a `let` var will have rspec cache that across calls and will break the impl.

PR #28, opened 2017-09-27
Gruf 2.0

splittingred:
## What? Why?

Gruf 2.0 is a *major* version bump, and has many backwards-incompatible changes. 

This version:
* Moves away from writing and declaring method implementations in the generated server stub, and now allows you to "bind" a Controller to a service to give you a pristine class to work on
* Restricts various APIs that are no longer needed
* Adds a cleaner API for developing RPC methods in a `Gruf::Controllers::Base` class and sets up framework for future feature development
* Adds more inject-ability into the `Gruf::Server` class
* Removes hooks. They are now replaced by a pure `Gruf::Interceptors::ServerInterceptor` class, which paves the way for:
  * Deprecating the arbitrary difference between authentication, instrumentation, and generic hooks
  * Delegated complete control of the order of execution of various interceptors
  * A more controlled API for interceptors that has a view of a controller request "context"

## New Controllers

Controllers now bind to services:

```ruby
class ThingController < ::Gruf::Controllers::Base
  bind ::Rpc::ThingService::Service

  def get_thing
    thing =  Rpc::Thing.new(id: request.message.id, name: 'Foo')
    Rpc::GetThingResponse.new(thing: thing)
  end
end
```

You'll also notice a new `request` object that contains the `message` object, as well as other things (the service class, active call, metadata, etc).

These controllers are also fully thread-safe, as opposed to the old `Gruf::Service`, which was shared between calls.

## How was it tested?

* `rspec`
* `spec/demo_server` then executing various rake commands
* Using in [gruf-demo](https://github.com/bigcommerce/gruf-demo) - see [releases/2.x branch](https://github.com/bigcommerce/gruf-demo/tree/releases/2.x)


README.md
```diff
@@ -8,17 +8,18 @@ provide a more streamlined integration into Ruby and Ruby on Rails applications.
...
-* Error data serialization in output metadata to allow fine-grained error handling in the transport while still 
-preserving gRPC BadStatus codes
+* Error data serialization in output metadata to allow fine-grained error handling in the transport while 
+  still preserving gRPC BadStatus codes
 * Server and client execution timings in responses
 
-gruf currently has active support for gRPC 1.4.x. gruf is compatible and tested with with Ruby 2.2, 2.3, and 2.4. gruf 
-is also not [Rails](https://github.com/rails/rails)-specific, and can be used in any Ruby framework (such as 
-[Grape](https://github.com/ruby-grape/grape), for instance).
+gruf currently has active support for gRPC 1.4.x-1.6x. gruf is compatible and tested with with Ruby 2.2,
```
junedkazi: 2 `with`

junedkazi: should it be `1.6.x` ?

README.md
```diff
@@ -308,7 +269,7 @@ It supports formatters (including custom ones) that you can use to specify the f
  
 ```ruby
 Gruf.configure do |c|
-  c.instrumentation_options[:request_logging] = {
+  c.interceptors[Gruf::Interceptors::Instrumentation::RequestLogging::Interceptor] = {
```
billthompson: This style of registering interceptors feels like it gives a little too much power to the consumers. From this (I haven't looked at the implementation yet), I think is probably using a hash to guarantee uniqueness. The problem with this approach is when interceptor order becomes important.

I think we can get more flexibility to change and new features by following a process like Rack::Middleware does:
```
c.interceptors.use Gruf::Interceptors::Instrumentation::RequestLogging::Interceptor, options: { formatter: :logstash}
```

Then, in the future, we can add the ability to control order without changing the underlying data structure via something like:
```
c.interceptors.insert_before ...
c.interceptors.insert_after ...
```

What do you think?

splittingred: I thought about this. So in Ruby insertion order is maintained in Hash ([see here](https://www.igvita.com/2009/02/04/ruby-19-internals-ordered-hash/)).

I understand the idea here though, and agree with you that implicit behavior can be a bit confusing. I'll look into this.

spec/pb/thing_controller.rb
```diff
@@ -15,65 +15,68 @@
 # OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #
 require 'rpc/ThingService_services_pb'
+require 'rpc/Error_pb'
 require 'ffaker'
 
-class ThingService < ::Rpc::ThingService::Service
-  include Gruf::Service
+class ThingController < ::Gruf::Controllers::Base
```
billthompson: Can we provide a more concrete example in the specs and documents? I feel like `Thing*` is just too abstract to easily comprehend. I'd like to see us use `User*`, as most developers can relate to that or have built systems with users. While the code will be the same, I think it will be easier to apply to the developer's situation using a less abstract concept.

README.md
billthompson: Awesome! Thanks for providing a link on hash insertion order and looking into changing the API.

UPGRADING.md
```diff
@@ -1,5 +1,117 @@
...
+Before, after, around, and outer around hooks have been removed in favor of the new 
+`Gruf::Interceptors::ServerInterceptor`. This paves the way for Gruf to support Client interceptors
+and also support the native gRPC interceptors added in [gRPC 1.7](https://github.com/grpc/grpc/pull/12100).
+
+Furthermore, Gruf 1.x had three different types of hooks with slightly different signatures: Authentication,
+Instrumentation, and generic hooks. Those have all been consolidated into interceptors.
+
+Interceptors behave very similarly to outer around hooks, with the following changes:
+
+* They are now provided an instance of `Gruf::Controllers::Request`, which contains request information
```
billthompson: 🌟

bin/gruf
```diff
@@ -17,13 +17,20 @@
...
 begin
-  server = Gruf::Server.new
+  server = Gruf::Server.new(Gruf.server_options)
+  Gruf.services.each { |s| server.add_service(s) }
+  Gruf.interceptors.each { |i, opts| server.add_interceptor(i, opts) }
   server.start!
 rescue => e
-  Gruf.logger.fatal "FATAL ERROR: #{e.message} #{e.backtrace.join("\n")}"
+  msg = "FATAL ERROR: #{e.message} #{e.backtrace.join("\n")}"
+  if Gruf.logger
```
billthompson: Can Gruf.logger be a `Logger.new(STDOUT)` if one is not provided during configuration? That would allow us to omit this conditional here and simplify logging in general.

lib/gruf/configuration.rb
```diff
@@ -85,27 +82,24 @@ def reset
       VALID_CONFIG_KEYS.each do |k, v|
         send((k.to_s + '='), v)
       end
+      self.root_path = Rails.root.to_s.chomp('/') if defined?(Rails)
```
billthompson: 🍺 It would probably better to have a `gruf-rails` gem to handle framework specific things like this in the future.

lib/gruf/controllers/base.rb
```diff
@@ -0,0 +1,103 @@
...
+      # @param [GRPC::GenericService] service
+      #
+      def self.bind(service)
+        self_klass = self
+        Gruf.services << service.name.constantize
+        service.rpc_descs.each do |name, rpc_desc|
+          method_key = name.to_s.underscore.to_sym
+
+          service.class_eval do
+            if rpc_desc.request_response?
```
billthompson: 🍺 It would be nice to encapsulate this logic into a `Gruf::Controllers::MethodFactory` (or something like that)

bin/gruf
splittingred: Yup, it is. `self.logger = ::Logger.new(STDOUT)` in configuration.rb.

However, this conditional here catches it on the chance it fatals before that logger is setup.

lib/gruf/controllers/request.rb
```diff
@@ -15,51 +15,75 @@
...
+        "#{service_key}.#{method_key}"
+      end
+
+      ##
+      # Return all messages for this request, properly handling different request types
+      #
+      # @return Enumerable<Object>
+      #
+      def messages
+        if type == :client_streamer
```
billthompson: 🍺 IMHO, this class is asking for a `RequestType` object to manage the logic of working with a specific request type. That could help this class offload the task conditional around L40 in `type` by building the `RequestType` object from the `rpc_desc` and also delegate the formatting of the messages to the RequestType class instead of doing that here.

I understand this may be considered as either an SRP violation or a premature abstraction so that's why I'm making this suggestion as optional.

If you need any clarification on this suggestion just let me know.

lib/gruf/errors/handlers.rb
```diff
@@ -0,0 +1,77 @@
...
+      ##
+      # Will issue a GRPC BadStatus exception, with a code based on the code passed.
+      #
+      # @param [Symbol] error_code The network error code that maps to gRPC status codes
+      # @param [Symbol] app_code The application-specific code for the error
+      # @param [String] message (Optional) A detail message about the error
+      # @param [Hash] metadata (Optional) Any metadata to inject into the trailing metadata for the response
+      #
+      def fail!(error_code, app_code = nil, message = '', metadata = {})
+        e = error
```
billthompson: 🍺 Build the error using a factory/builder pattern since there are transformations and conditional logic for the error state not specific to the handler logic.

lib/gruf/errors/handlers.rb
```diff
@@ -0,0 +1,77 @@
...
+# Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+# WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+# OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+#
+module Gruf
+  module Errors
+    module Handlers
```
billthompson: ⭐️ for making `fail` and `add_field_error` public 🙇

lib/gruf/errors/handlers.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      ##
+      # Add a field error to this endpoint
+      #
+      # @param [Symbol] field_name The name of the field
+      # @param [Symbol] error_code The application error code for the field
+      # @param [String] message A given error message for the field
+      #
+      def add_field_error(field_name, error_code, message = '')
+        error.add_field_error(field_name, error_code, message)
```
billthompson: This makes me feel like `@error` should be an instance variable on the controller instance and this module just exposes convenience methods for interacting with it. Not a blocker, but would be nice to have I think as the controller should be responsible for managing the error state. And then this is more of an `Errors::Helpers` mixin instead of a handler.

lib/gruf/interceptors/context.rb
```diff
@@ -15,30 +15,41 @@
...
-      # @return [Hash] The resulting output metadata with the timer attached
+      # @param [Gruf::Controllers::Request] request
+      # @param [Array<Gruf::Interceptors::ServerInterceptor>] interceptors
       #
-      def call(rc)
-        return unless rc.active_call && rc.active_call.respond_to?(:output_metadata)
-
-        rc.active_call.output_metadata.update(metadata_key => rc.execution_time.to_s)
+      def initialize(request, interceptors = [])
+        @interceptors = interceptors.map { |klass, opts| klass.new(request, opts) }
```
billthompson: 🍺 `Array(interceptors).map...`

lib/gruf/interceptors/instrumentation/output_metadata_timer.rb
```diff
@@ -15,29 +15,34 @@
...
-          # @param [Hash] payload The incoming request payload
-          # @return [String] The formatted string
-          #
-          def format(payload)
-            time = payload.fetch(:duration, 0)
-            grpc_status = payload.fetch(:grpc_status, 'GRPC::Ok')
-            route_key = payload.fetch(:method, 'unknown')
-            body = payload.fetch(:message, '')
+        def call
+          return unless request.active_call && request.active_call.respond_to?(:output_metadata)
```
billthompson: Since `nil` responds to `respond_to?`, you can get away with just `request.active_call.respond_to?(:output_metadata)` here.

lib/gruf/interceptors/instrumentation/output_metadata_timer.rb
```diff
@@ -15,29 +15,34 @@
...
 
-            msg = "[#{grpc_status}] (#{route_key}) [#{time}ms] #{body}".strip
-            msg += " Parameters: #{payload[:params].to_h}" if payload.key?(:params)
-            msg.strip
+          result = Gruf::Interceptors::Timer.time do
+            yield
           end
+          request.active_call.output_metadata.update(metadata_key => result.elapsed.to_s)
```
billthompson: 💁‍♂️ Law of Demeter here.

lib/gruf/interceptors/instrumentation/request_logging/interceptor.rb
```diff
@@ -0,0 +1,198 @@
...
+        #   Gruf::Instrumentation::Registry.add(:request_logging, Gruf::Instrumentation::RequestLogging::Hook)
+        #
+        class Interceptor < ::Gruf::Interceptors::ServerInterceptor
+
+          ###
+          # Log the request, sending it to the appropriate formatter
+          #
+          # @return [String]
+          #
+          def call
```
billthompson: 🍺 This method is has a high cyclomatic complexity. Consider breaking this up.

lib/gruf/interceptors/instrumentation/request_logging/interceptor.rb
```diff
@@ -0,0 +1,198 @@
...
+            if result.successful?
+              type = options.fetch(:success_log_level, :info).to_sym
+              status_name = 'GRPC::Ok'
+            else
+              type = options.fetch(:failure_log_level, :error).to_sym
+              status_name = result.message_class_name
+            end
+
+            payload = {}
+            if !client_streamer?
```
billthompson: If you perform the suggested refactor to build a `RequestType` object, you can update this part of the code to as the `RequestType` directly if it is a `client_streamer?` instead of using the adhoc logic on L89.

lib/gruf/server.rb
```diff
@@ -64,44 +64,54 @@ def start!
...
     # Auto-load all gRPC handlers
     #
     # :nocov:
     def setup!
-      Dir["#{Gruf.servers_path}/**/*.rb"].each do |f|
+      return unless File.directory?(controllers_path)
+      path = File.realpath(controllers_path)
+      $LOAD_PATH.unshift(path)
+      Dir["#{path}/**/*.rb"].each do |f|
+        next if f.include?('_pb') # exclude if people include proto generated files in app/rpc
```
billthompson: That is a pretty opinionated assumption based on naming convention - is it necessary? I'd prefer instead if we would inspect the file contents and only require it is it subclasses `::Gruf::Controllers::Base`. That would also protect us if someone includes controller concerns in the controller's directory - (see https://github.com/bigcommerce/gruf/pull/25)

lib/gruf/errors/handlers.rb
splittingred: To do this, I'm going to actually pass that instance variable into the interceptor initializers as well, so interceptors will be able to modify error state. I think this is useful for a few reasons - allowing interceptors to "fail gracefully" and handle multiple interceptors failing but not exiting early; also to allow field-level validation to move upstream without necessarily breaking out of the request cycle prior to the main method call.

billthompson: That's extremely powerful! Good call.

lib/gruf/server.rb
splittingred: Yes - due to the way gRPC generates and includes files, we need to ensure this file is not included (otherwise it will error out with a "Service already defined" error). Unfortunately this is an issue due to the way gRPC generates the Ruby files, and not something in Gruf.

spec/pb/thing_controller.rb
splittingred: Pushing to gruf 2.1

lib/gruf/interceptors/instrumentation/request_logging/interceptor.rb
splittingred: Pushing to gruf 2.1

lib/gruf/errors/handlers.rb
splittingred: Pushing to 2.1, as this doesn't affect any public contracts, and I'm not sure we want to do this as it's just a field mapping

splittingred:
@billthompson Thanks for the feedback! Addressed your comments: Cleaned up the error interfaces functionality, added a `Gruf::Controllers::Request::Type` class, and made a new `Gruf::Controllers::ServiceBinder` class for binding a controller to a service.

lib/gruf/interceptors/base.rb
```diff
@@ -0,0 +1,51 @@
...
+module Gruf
+  module Interceptors
+    ##
+    # Base class for interception requests
+    #
+    class Base
+      # @var [Gruf::Controllers::Request] request
+      attr_reader :request
+      # @var [Gruf::Error] errors
+      attr_reader :errors
```
billthompson: 🍺 With a plural name, I would expect this to be a collection. What do you think about calling this just `error`?

lib/gruf/interceptors/registry.rb
```diff
@@ -0,0 +1,137 @@
...
+
+      private
+
+      ##
+      # Handle mutations to the interceptor registry in a thread-safe manner
+      #
+      def interceptors_mutex(&block)
+        @interceptors_mutex ||= begin
+          require 'monitor'
+          Monitor.new
```
billthompson: TIL 🙇

lib/gruf/interceptors/registry.rb
```diff
@@ -0,0 +1,137 @@
...
+      # Insert an interceptor after another specified interceptor
+      #
+      # @param [Gruf::Interceptors::Base] after_class The interceptor to insert after
+      # @param [Gruf::Interceptors::Base] interceptor_class The class of the interceptor to add
+      # @param [Hash] options A hash of options to pass into the interceptor during initialization
+      #
+      def insert_after(after_class, interceptor_class, options = {})
+        interceptors_mutex do
+          pos = nil
+          @registry.each_with_index do |o, idx|
```
billthompson: 🍺 This code can be simplified in both of the insert methods:
```
pos = @registry.find_index { |interceptor| interceptor.fetch(:klass, '') == after_class.to_s }
```

lib/gruf/server.rb
billthompson: Can we verify based on the contents of the file instead? That will allow us to ensure we only require Gruf::Controllers. The protobuf files won't be subclassing `::Gruf::Controllers::Base` so this method is the safest IMHO. The current loader will still load non-service files (such as ActiveSupport::Concerns) as mentioned in the previous PR.

lib/gruf/interceptors/base.rb
splittingred: Yeah I had moved it to `errors` as it's really a payload of potential errors, but thinking about it more it _is_ a "holistic" error that to the end client is a single "error". Renaming.

lib/gruf/server.rb
splittingred: 1) This doesn't _run_ or mount anything in those, which the prior did; it just loads the files.
2) We _can't_ load the file due to the require calls in the generated files, and I don't particularly want to have to do file opening and string parsing on the file itself.

splittingred:
@billthompson 🙇 ♻️

lib/gruf/server.rb
billthompson: Ok. In our existing implementation I don't think we are using concerns anymore, but if this causes an issue I'll raise a new issue at that time.

PR #36, opened 2017-12-12
Expose request rpc_desc

Parad0X:
## What? Why?

Exposing `rpc_desc` would make it possible to access output message type RPC method is supposed to return. IE:

```ruby
# service HelloService {
#   rpc hello(HelloRequest) returns (HelloResponse) { }
# }

class Controller
  bind HelloService

  def hello
    request.message.class # => HelloRequest
    response_message_type # => HelloResponse
  end

  private

  def response_message_type
    request.rpc_desc.output
  end
end
```

## How was it tested?

Rspec [TODO]


Parad0X:
Closed in favor of #38

PR #38, opened 2017-12-15
Add service, request_class, and response_class to controller request object

### LINKED ISSUE #37: Expose @service from Gruf::Controllers::Request to interceptors
I would like to instrument Gruf endpoints with New Relic. For this I plan to create an interceptor which will utilize the `ControllerInstrumentation#perform_action_with_newrelic_trace`. http://www.rubydoc.info/github/newrelic/rpm/NewRelic%2FAgent%2FInstrumentation%2FControllerInstrumentation:perform_action_with_newrelic_trace

In order for the requests to show up correctly, I need to pass in the `class_name` of the controller / service. It seems like the `@service` is a natural fit for this use case.

https://github.com/bigcommerce/gruf/blob/master/lib/gruf/controllers/request.rb#L54

Herre is an example of the code I wrote which can instrument Gruf endpoints using newrelic, (maybe this is something which can also be brought into Gruf).

```rb
require 'new_relic/agent'

module Interceptors
  module Gruf
    ##
    # New Relic transaction tracing for Gruf endpoints
    #
    class NewRelicTracer < ::Gruf::Interceptors::ServerInterceptor
      include ::NewRelic::Agent::Instrumentation::ControllerInstrumentation

      def call
        opts = {
          category: :controller,
          class_name: request.instance_variable_get('@service'),
          name: request.method_key
        }

        perform_action_with_newrelic_trace(opts) do
          yield
        end
      end
    end
  end
end
```

I would also gladly open a pr to add an attr_reader for the service but figure we may want to have a different approach to expose these values.

splittingred:
## What? Why?

Fixes #36 and fixes #37 by exposing the following new attributes to the Controller Request object:

* `.service` - The Service class that is being executed against
* `.response_class` - The class of the response object that is expected for this request. This can be used for abstracting out response serialization.
* `.request_class` - The class of the incoming request object

@Parad0X @pedelman This should solve your use cases without having to expose the RPC descriptor to the public interfaces.

## How was it tested?

rspec, and through manual tests.


Parad0X:
Yeah works for me 👍

lib/gruf/controllers/request.rb
```diff
@@ -67,6 +70,20 @@ def service_key
         @service.name.underscore.tr('/', '.').gsub('.service', '')
       end
 
+      ##
+      # @return [Class]
```
Parad0X: To nit pick a little do you mean `@return [Object]`. Or is `[Class]` the preferred ruby way?

splittingred: So `Class` _is_ an `Object` in Ruby, which is why I'm specific here about the actual class being returned.

Fwiw, the ancestry chain is: `Class -> Module -> Object -> Kernel -> BasicObject`

Parad0X: That makes sense. Thanks! :shipit:

splittingred:
Thanks @Parad0X ! I'm going to make sure this fits @pedelman's use case, and then we'll get this integrated.

PR #39, opened 2017-12-15
Interceptor registry improvements

splittingred:
## What? Why?

* Add ability to list, clear, insert before, and insert after to a server's interceptor
registry
* Ensure interceptors and services cannot be adjusted on the server after it starts to 
prevent threading issues and other unwanted behavior

## How was it tested?

rspec, `spec/demo_server`, and other testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,12 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 
 ### Pending release
 
+- Add ability to list, clear, insert before, insert after, and remove to a server's interceptor
+registry
+- Ensure interceptors and services cannot be adjusted on the server after it starts to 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,8 +2,12 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 
 ### Pending release
 
+- Add ability to list, clear, insert before, and insert after to a server's interceptor
```
pedelman: I think it would be helpful to be able to remove an interceptor from the registry. Especially handy if one wants to remove one of the defaults without needing to clear other interceptors.

splittingred:
@pedelman Good callout - added! Also added to the README which interceptors are loaded by default.

PR #40, opened 2018-01-04
Update README.md and make you konw gruf quickly

pathbox:
Hi~
    Today I do some grpc example in Rails with gem `gruf`, `gruf` is cool, but I had a hard time at first, because I can't find a nice example about `gruf`. Then I find `gruf-demo`, but I think it is not enough. At last, I make it and do a simple example about `gruf`, so I update the README.md that I hope others don't do the wrong things that I did with `gruf` .
Please review my pull requests

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,17 +8,17 @@ provide a more streamlined integration into Ruby and Ruby on Rails applications.
 It provides an abstracted server and client for gRPC services, along with other tools to help get gRPC services in Ruby
 up fast and efficiently at scale. Some of its features include:
 
-* Abstracted controllers with request context support 
-* Full interceptors with timing and unified request context support 
+* Abstracted controllers with request context support
+* Full interceptors with timing and unified request context support
 * Robust client error handling and metadata transport abilities
 * Server authentication via interceptors, with basic auth with multiple key support built in
 * TLS support for client-server auth, though we recommend using [linkerd](https://linkerd.io/) instead
-* Error data serialization in output metadata to allow fine-grained error handling in the transport while 
+* Error data serialization in output metadata to allow fine-grained error handling in the transport while
   still preserving gRPC BadStatus codes
 * Server and client execution timings in responses
 
 gruf currently has active support for gRPC 1.4.x-1.8.x. gruf is compatible and tested with Ruby 2.2,
-2.3, and 2.4. gruf is also not [Rails](https://github.com/rails/rails)-specific, and can be used in any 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,11 +33,27 @@ Then in an initializer or before use:
...
+
+```ruby
+require 'gruf'
+
+Gruf.configure do |c|
+  c.default_client_host = 'grpc.service.com:9003'
+end
+```
+
+If you don't config `default_client_host`, you need do like this:
```
splittingred: Can you say: `If you don't explicitly set `default_client_host`, you will need to pass it into the client options, like so:`

README.md
```diff
@@ -115,29 +131,50 @@ Finally, you can start the server by running:
...
 
-Supporting an array of credentials allow for unique credentials per service, or for easy credential 
+In Client:
+
+```ruby
+require 'gruf'
+
+id = args[:id].to_i.presence || 1
+
+options = {username: 'my-username-here', password: 'my-password-here'}
```
splittingred: In order to promote best practices, let's do something like this here:

```ruby
options = {
  username: ENV.fetch('DEMO_THING_SERVICE_USERNAME'),
  password: ENV.fetch('DEMO_THING_SERVICE_PASSWORD')
}
```

README.md
```diff
@@ -301,26 +338,28 @@ It comes with a few more options as well:
...
 reports for gruf services
 
 ## Demo Rails App
 
-There is a [demonstration Rails application here](https://github.com/bigcommerce/gruf-demo) you can 
-view and clone that shows how to integrate Gruf into an existing Rails application. 
+There is a [demonstration Rails application here](https://github.com/bigcommerce/gruf-demo) you can
+view and clone that shows how to integrate Gruf into an existing Rails application.
+
+And another demo [Rails_Communicate_Rails](https://github.com/pathbox/Rails_Communicate_Rails) This demo has tow Rails App, one is grpc client, another is grpc server. It is "say hello" example. It is very simple, but it make you know `gruf` quickly.
```
splittingred: Thanks for creating this! Though, I'd rather keep the referenced demo apps in the `gruf` organization (we're looking at moving gruf to its own org in the next year) just so that we can make sure they're up-to-date and kept in line with best practices.

For now, can you remove this? We may create a section in the Wiki later that has "3rd Party examples" and such that would be more appropriate here.

pathbox:
Thank you for reviewing my request, I will do some modifications.

PR #42, opened 2018-02-01
Add ability to pass in client stub options to Gruf::Client

splittingred:
## What? Why?

Currently there is no way to pass in options into the underlying client stub when creating a `Gruf::Client`. This adds the ability to do so, which can be used for passing things like a default timeout, propagation mask, core interceptors, and channel arguments.

## How was it tested?

rspec, and through console testing.

----

@pedelman @billthompson @mattolson @zackangelo


lib/gruf/client.rb
```diff
@@ -66,14 +66,15 @@ def initialize(error)
...
     #
-    def initialize(service:, options: {})
+    def initialize(service:, options: {}, client_options: {})
       @base_klass = service
       @service_klass = "#{base_klass}::Service".constantize
       @opts = options || {}
       @opts[:password] = options.fetch(:password, '').to_s
       @opts[:hostname] = options.fetch(:hostname, Gruf.default_client_host)
-      client = "#{service}::Stub".constantize.new(@opts[:hostname], build_ssl_credentials)
+      client = "#{service}::Stub".constantize.new(@opts[:hostname], build_ssl_credentials, client_options)
```
billthompson: Can you provide / point me to more info on this `::Stub` namespace and how that works? Seeing that we have to  define a `timeout` method to access this value I'm concerned that we would need to define other methods to access the values that might also be passed into the `client_options` (which is not ideal).

Maybe it would be helpful to have a test that passes a `client_options` and not have a method explicitly defined to confirm this behavior?

splittingred: @billthompson It's the underlying client stub https://github.com/grpc/grpc/blob/master/src/ruby/lib/grpc/generic/client_stub.rb

And yes, there is a test I added in this PR that covers the timeout accessor.

I don't want to expose any other options; only the default timeout for referencing.

billthompson: Awesome thanks!

To clarify, you want the ability to specify `propagation mask, core interceptors, and channel arguments` in the `client_options` hash, but you dont' want the caller to be able to ask the client for those values? If so, can you elaborate on the vision here.

If that's not the case and you only want to set and retrieve the timeout option, can you update the PR description?

splittingred: > To clarify, you want the ability to specify propagation mask, core interceptors, and channel arguments in the client_options hash, but you dont' want the caller to be able to ask the client for those values? If so, can you elaborate on the vision here.

No, I do not. They are not modifiable after instantiation and the gRPC core does not expose them (so that they can remain immutable). I'd prefer to keep in line with gRPC core there.

With regard to timeout, that value is frozen and can safely be exposed.

The PR description still applies; if you look at the initializer, you can see I'm allowing _setting_ of all the options, but only _exposing_ an accessor for timeout.

PR #43, opened 2018-02-18
Trap INT/TERM signals and stop gracefully

Parad0X:
## What? Why?

grpc starting with 1.8.0 expects signals to be handled slightly differently. Running and stopping server in the same thread now crashes with:

```
E0218 00:01:31.158929000 140736335635392 completion_queue.cc:490]      assertion failed: cqd->completed_head.next == (uintptr_t)&cqd->completed_head
Abort trap: 6
```

Also `run_till_terminated` is an alias for `run`

Stopping server in a separate thread fixes this issue. 

```
Booting gRPC Server...
handling /grpc.health.v1.Health/Check with #<Method: Grpc::Health::V1::Health::Service#check>
calling 127.0.0.1:50051:/grpc.health.v1.Health/Check
Shutting down...
stopping, will wait for all the workers to exit
stopped: #<GRPC::RpcServer:0x007f98fe17c2a0>
stopped, all workers are shutdown
Goodbye!
```

## How was it tested?

Monkeypatched code has been running in production for a few days with regular deploys. No issues so far.

### DIFF
--- a/lib/gruf/server.rb
+++ b/lib/gruf/server.rb
@@ -39,7 +39,10 @@ module Gruf
       @interceptors = Gruf::Interceptors::Registry.new unless @interceptors.is_a?(Gruf::Interceptors::Registry)
       @services = []
       @started = false
-      setup!
+      @stop_server = false
+      @stop_server_cv = ConditionVariable.new
+      @stop_server_mu = Monitor.new
+      setup
     end
 
     ##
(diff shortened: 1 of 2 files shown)


splittingred:
This seems related to https://github.com/grpc/grpc/issues/14043 - I'd like to do some research here. I'm not sure just dropping it in a Thread without a proper monitor will solve this right.

Thanks for raising this, though, @Parad0X !

Parad0X:
@splittingred as long as start/stop are called in separate threads the issue is fixed but I updated my code to trap signals instead of rescuing `SignalException`.

gruf.gemspec
```diff
@@ -36,7 +36,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rake', '~> 10.0'
   spec.add_development_dependency 'pry'
 
-  spec.add_runtime_dependency 'grpc', '~> 1.4'
-  spec.add_runtime_dependency 'grpc-tools', '~> 1.4'
+  spec.add_runtime_dependency 'grpc', '~> 1.9'
```
splittingred: Question here - do we _need_ to bump the minimum dependency to 1.9? Can this work with 1.4?

I'd hate to force a downstream upgrade for users if there's not an explicit reason to do so.

Parad0X: Honestly probably not. Let me check if updated signal handling code works with 1.4, 1.5 and 1.7

Parad0X: Seems to be working fine with 1.4. I can revert.

Parad0X: On the other hand `1.4` is 9 months old and there has been a ton of major improvement/bug fixes in grpc ruby code (memory leak fixes, segfault fixes and what not) so I think its a good idea to encourage people to update.

splittingred: Understood, but I don't want to discourage usage because of an unnecessary version requirement. I think it's probably best to leave at 1.4 until we have need to bump it higher. Because it's loosely tied, people are free to push it higher downstream.

Parad0X: Additionally `1.9.0` adds ruby 2.5 support.

Parad0X:
@splittingred sounds good. Updated.

splittingred:
@Parad0X Can you squash your commits?

PR #44, opened 2018-02-24
Change proc title once server is ready to process incoming requests

Parad0X:
## What? Why?

Updating proc title is an alternative to https://github.com/grpc/grpc/blob/master/doc/health-checking.md  if you don't want every gruf app to implement this health checking service.

With this change it would be possible to run `ps | grep [g]ruf.*serving` to determine if gruf is ready to receive requests.

This change is also backwards compatible with `ps | grep gruf` or `pgrep gruf` if anyone was doing that to check whether gruf is up and running.

## How was it tested?

```shell
> bundle exec gruf 
Booting gRPC Server...
handling /grpc.health.v1.Health/Check with #<Method: Grpc::Health::V1::Health::Service#check>
# Ctrl+C
Shutting down...
stopping, will wait for all the workers to exit
stopped: #<GRPC::RpcServer:0x007fe70a0633f0>
stopped, all workers are shutdown
Goodbye!
```

```shell
> ps | grep gruf
14100 ttys003    0:06.11 gruf 2.2.0 -- starting
14100 ttys003    0:06.16 gruf 2.2.0 -- serving
14100 ttys003    0:06.18 gruf 2.2.0 -- stopped
```


Parad0X:
@splittingred fixed

Parad0X:
https://travis-ci.org/bigcommerce/gruf/jobs/346424539 

There's a race somewhere. Will ping you when I've found it.

Parad0X:
@splittingred looks like it's a pre-existing condition. I was able to reproduce it with `2.2.0`. 

```
+ git checkout 3c62904d4a5877a54240ff28a645f22709e594d6
+ rspec --format documentation

Gruf::Error
  functional test
    with a call that returns a field error
      should raise a Gruf::Client::Error and return the unserialized error object
    with multiple calls, where the first returns a field error, the second does not
      should not raise a Gruf::Client::Error the second time
    with multiple calls, with multiple having field errors

E0228 15:46:35.598241507   20673 completion_queue.cc:490]    assertion failed: cqd->completed_head.next == (uintptr_t)&cqd->completed_head
```

Could be a `grpc` bug. Don't have too much free time on my hands right now to investigate further but I'll try to look into it further later. We haven't seen any grpc segfaults in production so it could be just a race condition in specs.

How do you feel about merging this PR?

splittingred:
@Parad0X Hm, I thought this was fixed. I'll take a look at this this weekend and early next week (apologies, was busy with other work this week).

Thanks for looking into this. I'm pretty sure this is all related to: https://github.com/grpc/grpc/issues/14043 I'd like to be careful about embedding the processes into too many monitored threads because of the inherent complexity when doing that. That said, I think we can find a solution here. Thanks!

Parad0X:
Looks like it was just fixed again!

https://github.com/grpc/grpc/pull/14134

Parad0X:
I have been unable to reproduce this segfault with `grpc=1.10.0`

splittingred:
@Parad0X Cool. In our testing I'm seeing a lot less issues with 1.10 as well. LGTM - thanks for your due diligence on testing this!

splittingred:
@Parad0X https://github.com/bigcommerce/gruf/releases/tag/v2.2.1 https://rubygems.org/gems/gruf/versions/2.2.1

Parad0X:
Thanks @splittingred

PR #47, opened 2018-03-30
add `--pid-file` `--daemon` options when start gruf

NaixSpirit:
## What? Why?
add `--pidfile` and `--daemon` options, so we can use bundle exec grpc -d -P [YOUR_PID_FILE] to make the gruf server daemonized.

## How was it tested?

rspec

### DIFF
--- a/lib/gruf/cli/executor.rb
+++ b/lib/gruf/cli/executor.rb
@@ -21,6 +21,8 @@ module Gruf
     # Handles execution of the gruf binstub, along with command-line arguments
     #
     class Executor
+      attr_reader :opts
+
       ##
       # @param [Hash|ARGV]
       #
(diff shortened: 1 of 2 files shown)


lib/gruf/cli/executor.rb
```diff
@@ -62,9 +72,40 @@ def setup!
...
-        Gruf.server_binding_url = opts[:host] if opts[:host]
-        Gruf.use_default_interceptors = false if opts.suppress_default_interceptors?
-        Gruf.backtrace_on_error = true if opts.backtrace_on_error?
+        Gruf.server_binding_url = @opts[:host] if @opts[:host]
+        Gruf.use_default_interceptors = false if @opts.suppress_default_interceptors?
+        Gruf.backtrace_on_error = true if @opts.backtrace_on_error?
+      end
+
+      def daemonize
+        return if !@opts.daemon?
```
splittingred: It'd be better if we just didn't call this method if daemonizing isn't passed

NaixSpirit:
Thanks for your reply and sorry for my poor English.
Your suggestion is correct, Basically the intention is to remove the service management logic from the container and have it as part of the infrastructure

Issue #49, opened 2018-05-11
Gruf logs `Grpc::InvalidArgument` as an error rather than info


pedelman:
When utilizing the Gruf request logger, when the service raises a `Grpc::InvalidArgument`, Gruf is marking that as an `error` severity.

I consider `InvalidArgument` to be more of an "allowed" or "expected" behavior, and shouldn't be logged at such a severity. I propose we would handle this error as `info` level.

```rb
fail!(:bad_request, :bad_request, 'Bad Request')
```

Code that determine the log severity: [lib/gruf/interceptors/instrumentation/request_logging/interceptor.rb#L52-L58](https://github.com/bigcommerce/gruf/blob/master/lib/gruf/interceptors/instrumentation/request_logging/interceptor.rb#L52-L58)

splittingred:
Fixed in #51, released in 2.4.0

PR #51, opened 2018-05-24
fix(platform): GRUF-49 Interpret InvalidArguments as info, not errors.

gbanuel:
## What? Why?
Fixes issue #49, returns as INFO for InvalidArguments

## How was it tested?
See spec. Assume we return as INFO if we raise InvalidArgument exception.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 
 ### 2.4.0
 
-- Added three log levels: warn, info, error, mapped to explicit sets of GRPC error codes [#49]
-- Override log levels supplying info_level_codes, warn_level_codes, and error_level_codes in options of Interceptor constructor
-- Non-explicitly mapped error codes will be treated as :error level
+- Added a hash of error log levels to Interceptor class, mapping error code to level of logging to use.
+- To override the level of logging per error response, provide a map of codes to log level in options, key :log_levels. 
+- Default is :error log level.
 
 ### 2.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 
 ### Pending release
 
+### 2.4.0
+
+- Added three log levels: warn, info, error, mapped to explicit sets of GRPC error codes [#49]
+- Override log levels supplying info_level_codes, warn_level_codes, and error_level_codes in options of Interceptor constructor
+- Non-explicitly mapped error codes will be treated as :error level
+
 ### 2.3.0
 
 - Add Gruf::Interceptors::ClientInterceptor for intercepting outbound client calls
--- a/lib/gruf/interceptors/instrumentation/request_logging/interceptor.rb
+++ b/lib/gruf/interceptors/instrumentation/request_logging/interceptor.rb
@@ -14,7 +14,6 @@
 # OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #
 require 'socket'
-require 'pry'
 require_relative 'formatters/base'
 require_relative 'formatters/logstash'
 require_relative 'formatters/plain'
--- a/lib/gruf/interceptors/instrumentation/request_logging/interceptor.rb
+++ b/lib/gruf/interceptors/instrumentation/request_logging/interceptor.rb
@@ -14,6 +14,7 @@
 # OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #
 require 'socket'
+require 'pry'
 require_relative 'formatters/base'
 require_relative 'formatters/logstash'
 require_relative 'formatters/plain'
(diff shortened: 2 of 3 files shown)


zvuki:
How about other code?
```
The following status codes are never generated by grpc library:

INVALID_ARGUMENT
NOT_FOUND
ALREADY_EXISTS
FAILED_PRECONDITION
ABORTED
OUT_OF_RANGE
DATA_LOSS
```

splittingred:
@zvuki Technically those _are_ unsuccessful requests, IMO, whereas `InvalidArgument` usually implies an invalid input on behalf of the user through a validation.

zvuki:
@splittingred if you consider them in terms of http response codes, I think they all fall into 4XX - client-side errors

splittingred:
Do you all think we should log those codes as INFO then?

- INVALID_ARGUMENT - I think this is OK to, as it's validation-centric
- NOT_FOUND - Generally this is accepted as an actual error (404), but I could see an argument that this is expected behavior for the service.
- ALREADY_EXISTS - I could see this being perceived as an error (conflict) or not (uniqueness validation)
- FAILED_PRECONDITION - This is a server error, usually dealing with system state, and IMO should be logged as ERROR
- ABORTED - This is a concurrency/network error and should be ERROR
- OUT_OF_RANGE - This is generally an error dealing with pagination from a client request. I could see this either way.
- DATA_LOSS - This is a server/network error and should be logged as ERROR

So the following to me need discussion:
- INVALID_ARGUMENT
- NOT_FOUND
- ALREADY_EXISTS
- OUT_OF_RANGE

Thoughts? @pedelman @zvuki @gbanuel

pedelman:
Imo since it seems like there will be some subjectivity, maybe we should just allow the user to define for each grpc error type, what log level they prefer.

Otherwise maybe something like;

INFO

```
INVALID_ARGUMENT
NOT_FOUND
ALREADY_EXISTS
OUT_OF_RANGE
```

ERROR

```
DATA_LOSS
ABORTED
FAILED_PRECONDITION
```

splittingred:
I'm wondering if there's a way we can make this configurable instead. @gbanuel - why don't we create an option for the request logging interceptor to pass in an array of grpc status code classes that it considers info-level. We can default to this setup (using [this list for reference](https://github.com/grpc/grpc-go/blob/master/codes/codes.go)):

```ruby
{
  error_codes: [GRPC::Unknown, GRPC::Internal, GRPC::DataLoss, GRPC::FailedPrecondition, GRPC::Unavailable, GRPC::DeadlineExceeded, GRPC:: Canceled],
  info_codes: [GRPC::Ok, GRPC::InvalidArgument, GRPC::NotFound, GRPC::AlreadyExists, GRPC::OutOfRange],
  warn_codes: [GRPC::Unauthenticated, GRPC::PermissionDenied]
}
```

I think we fallback to ERROR if the status code isn't mapped.
And then just check to see if the resulting status is one of those, and log the appropriate level. Thoughts all?

gbanuel:
Sounds good to me. The interceptor constructor will have an option to override the error map that will be part of the Error class. If not included, the Error class defined mapping will be followed as described above.

tpietsch:
maybe log metrics on the client OR server on what status codes are being sent that way large swings can be easily seen. If a deploy causes tons of not-founds/invalid arg something is prob broken and should be alerted on or otherwise.

splittingred:
> maybe log metrics on the client AND/OR server on what status codes are being sent that way large swings can be easily seen

FYI, this is done through the StatsD interceptor - so no need to do that here.

lib/gruf/interceptors/instrumentation/request_logging/interceptor.rb
```diff
@@ -49,12 +56,20 @@ def call
...
-            if result.successful?
-              type = options.fetch(:success_log_level, :info).to_sym
+            # Fetch from options if they have been supplied else default to internally defined mappings...
+            info_level_map = options.fetch(:info_level_codes, LOG_LEVEL_MAPS[:info_level_codes])
+            warn_level_map = options.fetch(:warn_level_codes, LOG_LEVEL_MAPS[:warn_level_codes])
+            error_level_map = options.fetch(:error_level_codes, LOG_LEVEL_MAPS[:error_level_codes])
+
+            type = :error
+            status_name = result.message_class_name
+            if result.successful? || info_level_map.include?(result.message.class)
```
pedelman: Any reason to check `successful?` here since we can just check if the response status is in the info_level_map?

lib/gruf/interceptors/instrumentation/request_logging/interceptor.rb
```diff
@@ -49,12 +56,20 @@ def call
...
+              type = :info
               status_name = 'GRPC::Ok'
```
pedelman: Do we still need this line?

lib/gruf/interceptors/instrumentation/request_logging/interceptor.rb
gbanuel: I was trying to prevent any code depending on that from failing. The successful flag previously was the only toggle between :error and :info. Now there's :warn. Anyone using this code as before should see it behave the same. That is, any successful result, regardless of specific code will still work as before. Since, however, we are augmenting the version, we can ignore it and make sure to note that in the change log, and just go by the response status. I'll make the correction.

lib/gruf/interceptors/instrumentation/request_logging/interceptor.rb
```diff
@@ -37,6 +37,13 @@ class InvalidFormatterError < StandardError; end
         #
         class Interceptor < ::Gruf::Interceptors::ServerInterceptor
 
+          # Default mappings of codes to log levels...
+          LOG_LEVEL_MAPS = {
```
pedelman: I wonder if it would be better to store a map of gRPC status code to log level. Right now in the code, we need to first create a mapping of the level to a list of status codes, this means whenever we want to look up a status code, we need to potentially loop over the entire list of status codes. For the case of looking up an error, I need to first loop over all the info, then all the warn, then (potentially) all the errors.

It also means that if I want to override the behavior of a single status code, I don't need to re-define ALL the status codes which might be in that group.

gbanuel: Correct on two fronts. I went with @splittingred 's suggestion of the hash of arrays. But this does force you to redefine them all to ensure that a code doesn't exist in multiple levels. A map per each is less elegant at first sight but a lot more efficient, both for lookup and replacement.

gbanuel:
I've left the check for "successful" as it will break tests to remove. Second, as I understand this code, an intercepted call can either be "successful" or some level of not. If yes, use OK and info, as this is what clients may expect. Everything else is some level of error.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 
 ### Pending release
 
+### 2.4.0
+
+- Added three log levels: warn, info, error, mapped to explicit sets of GRPC error codes [#49]
+- Override log levels supplying info_level_codes, warn_level_codes, and error_level_codes in options of Interceptor constructor
```
pedelman: I think this still needs updating

PR #52, opened 2018-06-08
Add logging for unhandled exceptions

mattolson:
## What? Why?
* We have good error handling in the request logging interceptor, but if an unhandled exception occurs, we don't currently log it.
* This adds logging with extended debug information, but preserves the behavior of what we send back to the client.

## How was it tested?
* Added a `fail` in one of my rpc handlers, saw the log message using the configured logger.

### DIFF
--- a/lib/gruf/controllers/base.rb
+++ b/lib/gruf/controllers/base.rb
@@ -73,6 +73,7 @@ module Gruf
       rescue GRPC::BadStatus
         raise # passthrough
       rescue StandardError => e
+        Gruf.logger.error "Unhandled #{e.class.name}: #{e.message} #{e.backtrace.join("\n")}"
         set_debug_info(e.message, e.backtrace) if Gruf.backtrace_on_error
         error_message = Gruf.use_exception_message ? e.message : Gruf.internal_error_message
         fail!(:internal, :unknown, error_message)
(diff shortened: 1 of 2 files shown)


lib/gruf/controllers/base.rb
```diff
@@ -73,6 +73,7 @@ def call(method_key, &block)
       rescue GRPC::BadStatus
         raise # passthrough
       rescue StandardError => e
+        Gruf.logger.error "Unhandled #{e.class.name}: #{e.message} #{e.backtrace.join("\n")}"
```
splittingred: So I think this should go into the RequestLogging interceptor, and not here. You likely just need to rescue and re-raise StandardError there.

splittingred: Also - you may want to truncate the backtrace to the last 5-8 lines, so it's not huge.

mattolson: This is a global exception handler for all code including any other interceptors that may be added, so I think it's useful to have it here...

splittingred: Right, but having it here makes it not easily configurable (as opposed to the interceptor) or overridable in how logs are made. I'd rather push as much logging to the interceptors themselves to allow them to be fully configurable.

mattolson: If I could guarantee that this error handler were the outermost interceptor, we could put all of this rescue logic there. What do you think about that? Create a new interceptor that is at the outermost layer, and move the begin...rescue logic there...

splittingred: @mattolson So you can do that today! Interceptor runtime is controlled entirely from configuration: https://github.com/bigcommerce/gruf#configuring-interceptors

They execute in a FIFO manner, so you can manage this as you want.

PR #54, opened 2018-07-31
Set exception message for Gruf::Client::Error

jasondoc3:
## What?

This PR sets the `message` attribute on instances of `Gruf::Client::Error` (when possible).

Before:

```rb
begin
  raise Gruf::Client::Error, { 'code' => 'internal', 'message' => 'An error occurred' }
rescue => e
  puts e.message
end

=> Gruf::Client::Error
```

After:
```rb
begin
  raise Gruf::Client::Error, { 'code' => 'internal', 'message' => 'An error occurred' }
rescue => e
  puts e.message
end

=> 'GRPC Internal: An error occurred'
```

## Why?

Error services (such as [Sentry](https://sentry.io/)) use the exception's message to more effectively group exceptions. If the message is always `Gruf::Client::Error`, different exceptions are improperly grouped together under the same event.

## How was it tested?

The specs + using a monkey patch on a service

### DIFF
--- a/lib/gruf/client.rb
+++ b/lib/gruf/client.rb
@@ -49,6 +49,23 @@ module Gruf
       #
       def initialize(error)
         @error = error
+        super(error_message)
+      end
+
+      private
+
+      ##
+      # Format the exception message
+      #
+      # @return [String] A formatted exception message
+      #
+      def error_message
+        if @error.is_a?(Hash) && @error.key?('code') && @error.key?('message')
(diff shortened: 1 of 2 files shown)


jasondoc3:
Thanks @splittingred for the detailed explanation and feedback. It makes a lot of sense! I will rescue `Gruf::Client::Error` in my application's code and re-raise a domain specific error + message.

PR #55, opened 2018-08-03
Safer configuration of RpcServer options

mattolson:
## What? Why?
* To more effectively separate Gruf::Server options from GRPC::RpcServer options,
  add a new configuration called rpc_server_options which contains just the known
  params for the RpcServer constructor.
* We still allow users to pass these in through Gruf.server_options, but that should
  be considered deprecated going forward.
* A side effect of this is that we are providing a proper default for pool_keep_alive
  which is set incorrectly by default in the grpc library. See https://github.com/grpc/grpc/pull/16240
  for details.

## How was it tested?
`bundle exec rspec`


lib/gruf/server.rb
```diff
@@ -31,10 +31,10 @@ class ServerAlreadyStartedError < StandardError; end
     ##
     # Initialize the server and load and setup the services
     #
-    # @param [Hash] options
+    # @param [Hash] opts
     #
-    def initialize(options = {})
-      @options = options || {}
+    def initialize(opts = {})
+      @options = opts || {}
```
splittingred: If you're going to change this name here, can you change to `@options` or `opts` on L46 below

mattolson:
I have updated the change log and the readme, but I don't think we need it in the upgrade guide because this change is backward compatible.

PR #56, opened 2018-08-07
PEN-855: Add monitoring of grpc server

mattolson:
## What? Why?
* GRPC::RpcServer does not expose events such as RESOURCE_EXHAUSTED
  or UNIMPLEMENTED to the interceptor chain or the application code
  other than through log messages.
* To allow for enhanced instrumentation by applications, especially
  in the case of thread pool exhaustion (which is critical to monitor),
  add a subclass that allows us to call an event listener when these
  or other critical events occur.
* Note that since we are digging into the internals a bit, we need
  to watch the evolution of the GRPC::RpcServer class. I tried to make
  this as light as possible, only working with methods that are unlikely
  to change.
* This subclass is only used if the caller sets `event_listener_proc`
  in the Gruf configuration.
* With this change, applications can instrument every case where the
  server rejects a request due to an exhausted thread pool, or when
  a method is unimplemented.

## How was it tested?
* Built an app that uses the local copy of gruf
* Added an event listener proc that increments a counter in statsd when the thread pool is exhausted
* Set the thread pool size to 1
* Fired off many concurrent requests and observed the events in statsd


splittingred:
This looks good.

Can you update the CHANGELOG?

PR #59, opened 2018-08-14
PEN-858: Handle GRPC::Core::CallError

mattolson:
## What? Why?
* Under load, and especially when thread pool is exhausted, the grpc library sometimes throws `GRPC::Core::CallError`. We should send a valid error response to the client in this case, rather than crash the process. We already have handling for this, but only if the exception inherits from `StandardError`, but `CallError` does not.
* Remove Thread.abort_on_exception. We should not crash the server if a thread has an unhandled exception. Rather, we should make sure our global error handler sends a valid error response to the client.

```
irb(main):004:0> GRPC::Core::CallError.is_a? StandardError
=> false
```

## References
* ["call#run_batch failed somehow"](https://github.com/grpc/grpc/blob/c50405364a194a0e4931153cbe329662d90530bc/src/ruby/ext/grpc/rb_call.c#L854)
* [PR where this was fixed for bidi but not unary calls](https://github.com/grpc/grpc/pull/14922/files)

## How was it tested?
* Built a grpc app with a simple health endpoint
* Wrote a client that sends traffic to that endpoint, with a deadline of 250ms
* Set `GRPC_SERVER_POOL_SIZE` to 1 to trigger thread pool exhaustion
* Observed `call#run_batch failed somehow` and process crash
* Ran again with these changes, and now we send a proper grpc error 13 response instead of crashing


splittingred:
:+1: Can you bump version to `2.4.2.pre` and also update the CHANGELOG.md?

lib/gruf/controllers/base.rb
```diff
@@ -71,8 +71,8 @@ def call(method_key, &block)
           send(method_key, &block)
         end
       rescue GRPC::BadStatus
-        raise # passthrough
-      rescue StandardError => e
+        raise # passthrough, to be caught by Gruf::Interceptors::Timer
+      rescue GRPC::Core::CallError, StandardError => e # CallError is not a StandardError
```
pedelman: Can we return `:grpc_call_error` as the second arg to the `fail!` method? Its not really an unknown error, it is a GRPC::CallError

mattolson: We would have to duplicate the error handling for CallError just for that, which I don't think is worth it. In this case, the error message that comes back should be sufficient to explain the error.

PR #60, opened 2018-08-17
PEN-925: Handle overflow of metadata

mattolson:
## What? Why?
* The gRPC C library has a default max size on trailing metadata of 8kb. Update the Error class to consider this when attaching metadata to the gRPC call. If the overall size of the metadata, including metadata added by the application, and the error payload added by the error serializer, is in danger of exceeding that limit, log the error and replace the wire error with a new one indicating metadata overflow.
* Truncate stack trace attached to error payload to ten lines to reduce risk of overflow.

## How was it tested?
* gRPC server app with an endpoint that calls `fail!(:internal, :unknown, SecureRandom.hex(9000))`
* Call it with a client that logs the response

Server side:
```
I, [2018-08-16T18:48:43.173244 #2735]  INFO -- : deadline is 1969-12-31 15:59:59 -0800; (now=2018-08-16 18:48:43 -0700)
I, [2018-08-16T18:48:43.173316 #2735]  INFO -- : schedule another job
W, [2018-08-16T18:48:43.392624 #2735]  WARN -- : metadata_size_exceeded: Metadata too long, risks exceeding http2 trailing metadata limit. Original error: {:code=>:internal, :app_code=>:unknown, :message=>"<big long random string>", :field_errors=>[], :debug_info=>{}}
D, [2018-08-16T18:48:43.394784 #2735] DEBUG -- : app err:#<GRPC::ActiveCall:0x007fb3c70c8fa0>, status:13:Metadata too long, risks exceeding http2 trailing metadata limit.
D, [2018-08-16T18:48:43.394822 #2735] DEBUG -- : Sending status  13:Metadata too long, risks exceeding http2 trailing metadata limit.
```

Client side:
```
D, [2018-08-16T18:48:41.600250 #2740] DEBUG -- : Failing with status #<struct Struct::Status code=13, details="Metadata too long, risks exceeding http2 trailing metadata limit.", metadata={"error-internal-bin"=>"\n\x16metadata_size_exceeded\x12AMetadata too long, risks exceeding http2 trailing metadata limit."}>
```

### DIFF
--- a/lib/gruf/error.rb
+++ b/lib/gruf/error.rb
@@ -50,6 +50,13 @@ module Gruf
       data_loss: GRPC::DataLoss
     }.freeze
 
+    # Default limit on trailing metadata is 8KB. We need to be careful
+    # not to overflow this limit, or the response message will never
+    # be sent. Instead, resource_exhausted will be thrown.
+    MAX_METADATA_SIZE = 7.5 * 1_024
+    METADATA_SIZE_EXCEEDED_CODE = 'metadata_size_exceeded'.freeze
+    METADATA_SIZE_EXCEEDED_MSG = 'Metadata too long, risks exceeding http2 trailing metadata limit.'.freeze
+
     # @return [Symbol] The given internal gRPC code for the error
     attr_accessor :code
     # @return [Symbol] An arbitrary application code that can be used for logical processing of the error by the client
@@ -58,8 +65,8 @@ module Gruf
     attr_accessor :message
     # @return [Array] An array of field errors that can be returned by the server
     attr_accessor :field_errors
-    # @return [Object] A hash of debugging information, such as a stack trace and exception name, that can be used to
-    # debug an given error response. This is sent by the server over the trailing metadata.
+    # @return [Errors::DebugInfo] A object containing debugging information, such as a stack trace and exception name,
+    # that can be used to debug an given error response. This is sent by the server over the trailing metadata.
     attr_accessor :debug_info
     # @return [GRPC::BadStatus] The gRPC BadStatus error object that was generated
     attr_writer :grpc_error
@@ -72,10 +79,11 @@ module Gruf
     # @param [Hash] args (Optional) An optional hash of arguments that will set fields on the error object
     #
     def initialize(args = {})
+      @field_errors = []
+      @metadata = {}
       args.each do |k, v|
         send("#{k}=", v) if respond_to?(k)
       end
-      @field_errors = []
     end
 
     ##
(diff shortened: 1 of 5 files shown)


zvuki:
Can it in addition to `Metadata too long` have at least a short exception message text of smth?

mattolson:
@zvuki While the error payload is the most likely culprit, it's also possible for the server to attach custom metadata to the error. So while we could try to repair just the error message, that is not enough in general to solve this problem. Rather than try to deeply inspect all the possible values and repair them, I opted for logging everything on the server and replacing the error on the wire with something simpler that is guaranteed not to cause issues.

lib/gruf/error.rb
```diff
@@ -50,6 +50,13 @@ class Error
       data_loss: GRPC::DataLoss
     }.freeze
 
+    # Default limit on trailing metadata is 8KB. We need to be careful
+    # not to overflow this limit, or the response message will never
+    # be sent. Instead, resource_exhausted will be thrown.
+    MAX_METADATA_SIZE = 7.5 * 1024
```
splittingred: `1_024`

lib/gruf/error.rb
```diff
@@ -130,16 +138,30 @@ def serialize
...
-    # @return [Error] Return the error itself with the GRPC::ActiveCall attached and error metadata appended
+    # @return [Error] Return the error itself after updating metadata on the given gRPC call.
+    #                 In the case of a metadata overflow error, we replace the current error with
+    #                 a new one that won't cause a low-level http2 error.
     #
     def attach_to_call(active_call)
       metadata[Gruf.error_metadata_key.to_sym] = serialize if Gruf.append_server_errors_to_trailing_metadata
-      if !metadata.empty? && active_call && active_call.respond_to?(:output_metadata)
-        active_call.output_metadata.update(metadata)
+      return self unless !metadata.empty? && active_call && active_call.respond_to?(:output_metadata)
```
splittingred: This is confusing to read. Prefer:

```ruby
return self if metadata.empty? || !active_call || !active_call.respond_to?(:output_metadata)
```

lib/gruf/error.rb
```diff
@@ -130,16 +138,30 @@ def serialize
...
+
+      # Check if we've overflown the maximum size of output metadata. If so,
+      # log a warning and replace the metadata with something smaller to avoid
+      # resource exhausted errors.
+      if metadata.inspect.size > MAX_METADATA_SIZE
+        code = METADATA_SIZE_EXCEEDED_CODE
+        msg = METADATA_SIZE_EXCEEDED_MSG
+        logger.warn "#{code}: #{msg} Original error: #{to_h.inspect}"
```
splittingred: Is there a chance we'd be leaking sensitive data here to logs?

lib/gruf/errors/debug_info.rb
```diff
@@ -30,7 +30,7 @@ class DebugInfo
       #
       def initialize(detail, stack_trace = [])
         @detail = detail
-        @stack_trace = stack_trace.is_a?(String) ? stack_trace.split("\n") : stack_trace
+        @stack_trace = (stack_trace.is_a?(String) ? stack_trace.split("\n") : stack_trace)[0..10]
```
splittingred: :+1:, though we should make this 10 limit configurable.

lib/gruf/error.rb
mattolson: Funny, I was going back and forth on this one. :)

lib/gruf/error.rb
mattolson: No more so than what would be sent back to the client in the error payload which is most likely logged anyway. I think the risk of additional exposure is minimal.

lib/gruf/error.rb
splittingred: Yeah, I waffle on these cases generally too, but have settled into what seems to be industry best practice, which is in a conditional, if there are multiple conditions, prefer `if` over `unless` as that way you're not dealing with double negatives.

lib/gruf/error.rb
splittingred: Okay, the difference there though is the client (and any interceptors) have control over what is logged. This way we're making it canon and unavoidable.

lib/gruf/errors/debug_info.rb
```diff
@@ -29,8 +29,10 @@ class DebugInfo
       # @param [Array<String>] stack_trace The stack trace generated by the exception as an array of strings
       #
       def initialize(detail, stack_trace = [])
+        stack_trace_limit = Gruf.backtrace_limit.to_i
+        stack_trace_limit = 10 if stack_trace_limit < 0
```
splittingred: You probably want to allow for a `0` limit, which would mean to not limit at all.

PR #61, opened 2018-08-20
Client exceptions raised now contain mapped subclasses; will handle non-badstatus errors

splittingred:
## What? Why?

- Client exceptions raised now contain mapped subclasses, such as `Gruf::Client::Errors::InvalidArgument`
- Client exceptions will also now catch StandardError and GRPC::Core errors, and handle them as Internal errors

These closely map with their `GRPC::BadStatus` counterparts, and each subclass `Gruf::Client::Error`. This should be fully backwards compatible with existing client error handling code, as the original exception through is still available via `.error` on the raised exception.

One change is that Gruf will now catch `StandardError` and `GRPC::Core::CallError` exceptions at the client boundary, and translate them into `Gruf::Client::Errors::Internal` exceptions in the client. For this reason, I've bumped the version to 2.5.0.pre.

## How was it tested?

`spec/demo_server` and rspec.

----

@bigcommerce/platform-engineering @mattolson @pedelman


lib/gruf/client/error.rb
```diff
@@ -0,0 +1,64 @@
...
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions of the
+# Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+# WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+# OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+#
+module Gruf
+  class Client < SimpleDelegator
```
mattolson: Is this meant to be a class, or a module?

lib/gruf/client/error_factory.rb
```diff
@@ -0,0 +1,99 @@
...
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions of the
+# Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+# WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+# OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+#
+module Gruf
+  class Client < SimpleDelegator
```
mattolson: Same here: Is this meant to be a class, or a module?

lib/gruf/client/error.rb
splittingred: Class. `Gruf::Client` is a class, and therefore cannot be redeclared as a module.

lib/gruf/client/error_factory.rb
splittingred: See above.

PR #62, opened 2018-09-12
Add SynchronizedClient to mitigate thundering herds

mattolson:
## What? Why?
To help with cases where you might have spikes of traffic to a given endpoint (so-called thundering herds), SynchronizedClient puts locks around Client calls to a given endpoint with the same params.

## How was it tested?
* Unit tests
* A server endpoint that uses Gruf::Client to call out to another service was updated to use Gruf::SynchronizedClient and a test client was used to generate load against it. I observed the debug logs and saw the expected behavior.

### DIFF
--- a/lib/gruf/synchronized_client.rb
+++ b/lib/gruf/synchronized_client.rb
@@ -1,0 +1,93 @@
+# Copyright (c) 2017-present, BigCommerce Pty. Ltd. All rights reserved
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+# documentation files (the "Software"), to deal in the Software without restriction, including without limitation the
+# rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit
+# persons to whom the Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions of the
+# Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+# WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+# OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+#
+require 'concurrent'
+
+module Gruf
+  ##
+  # Ensures that we only have one active call to a given endpoint with a given set of params. This can be useful
+  # to mitigate thundering herds.
+  #
+  class SynchronizedClient < Gruf::Client
+    attr_reader :unsynchronized_methods
+
+    ##
+    # Initialize the client and setup the stub
+    #
+    # @param [Module] service The namespace of the client Stub that is desired to load
+    # @param [Hash] options A hash of options for the client
+    # @option options [Array] :unsynchronized_methods A list of methods (as symbols) that should be excluded from synchronization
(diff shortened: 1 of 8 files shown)


lib/gruf/synchronized_client.rb
```diff
@@ -0,0 +1,102 @@
...
+    #
+    # @param [Module] service The namespace of the client Stub that is desired to load
+    # @param [Hash] options A hash of options for the client
+    # @option options [String] :password The password for basic authentication for the service.
+    # @option options [String] :hostname The hostname of the service. Defaults to linkerd.
+    # @option options [Float] :ttl How long to give until expiring results from previous calls (in seconds)
+    # @param [Hash] client_options A hash of options to pass to the gRPC client stub
+    #
+    def initialize(service:, options: {}, client_options: {})
+      @ttl = options[:ttl] || Gruf.client_cache_ttl
```
splittingred: `@ttl = options.fetch(:ttl, Gruf.client_cache_ttl)`

lib/gruf/synchronized_client.rb
```diff
@@ -0,0 +1,102 @@
...
+
+require 'concurrent'
+
+module Gruf
+  ##
+  # Ensures that we only have one active call to a given endpoint with a given set of params. This can be useful
+  # to mitigate thundering herds. Cache the result for a period of time so other requests, once unblocked, will get the
+  # same value. These results are expired after the given ttl.
+  #
+  class SynchronizedClient < Gruf::Client
```
splittingred: 💁‍♂️ This class needs tests.

lib/gruf/synchronized_client.rb
```diff
@@ -0,0 +1,102 @@
...
+    ##
+    # Initialize the client and setup the stub
+    #
+    # @param [Module] service The namespace of the client Stub that is desired to load
+    # @param [Hash] options A hash of options for the client
+    # @option options [String] :password The password for basic authentication for the service.
+    # @option options [String] :hostname The hostname of the service. Defaults to linkerd.
```
splittingred: I think you can remove the hostname/password options here from the docblock as their implied through inheritance

lib/gruf/synchronized_client.rb
```diff
@@ -0,0 +1,102 @@
...
+    # for the given above call
+    # @param [Hash] metadata (Optional) A hash of metadata key/values that are transported with the client request
+    # @param [Hash] opts (Optional) A hash of options to send to the gRPC request_response method
+    # @return [Gruf::Response] The response from the server
+    # @raise [Gruf::Client::Error|GRPC::BadStatus] If an error occurs, an exception will be raised according to the
+    # error type that was returned
+    #
+    def call(request_method, params = {}, metadata = {}, opts = {}, &block)
+      # Generate a unique cache key based on the method and params
+      cache_key = "#{request_method}.#{params.hash}"
```
splittingred: Q for you - can cache keys with Concurrent::Map have `.-` in them? (`.hash` can sometimes return a negative number)

lib/gruf/synchronized_client.rb
```diff
@@ -0,0 +1,102 @@
...
+      lock.synchronize do
+        # Get value from cache or create placeholder
+        response = @cache.get(cache_key)
+        if response
+          Gruf.logger.debug "Returning cached result for #{cache_key}"
+          next response
+        end
+
+        # Make the call and record it in the cache
+        response = super
```
splittingred: It might be worth it making a note as to what would happen should the client call here raise an exception

mattolson: I've already tested this, and `synchronize` will properly release the lock and the exception is raised to the caller.

lib/gruf/synchronized_client.rb
mattolson: Yeah, any string will do.

lib/gruf/synchronized_client.rb
```diff
@@ -0,0 +1,94 @@
...
+module Gruf
+  ##
+  # Ensures that we only have one active call to a given endpoint with a given set of params. This can be useful
+  # to mitigate thundering herds.
+  #
+  class SynchronizedClient < Gruf::Client
+    attr_reader :unsynchronized_methods
+
+    # How often we sweep the internal results cache (in seconds)
+    EXPIRY = 10
```
splittingred: We should make this configurable in `Gruf::Client`'s `options` argument

mattolson: This is really an internal concern of the class. The cache behavior does not affect callers -- it is only used to communicate results from the first thread to any others blocked on the same lock. However, for the sake of completeness, I have added a configuration for it.

PR #66, opened 2018-11-08
Log all server's errors in request logging

kruczjak:
## What? Why?

When there is unhandled server exception (e.g. NoMethodError) then no request log is logged.
Additionally, `Gruf::Interceptors::Timer` should count time always when there is error (not just `GRPC::BadStatus`)

## How was it tested?

Test :)


splittingred:
Closing this due to inactivity, please feel free to re-open after changes are made. Thanks!

PR #70, opened 2019-02-18
Log ok/validation responses at DEBUG levels to prevent log stampeding

splittingred:
## What? Why?

Reduces defaults for OK/validation-level errors in the request log handler to `debug`, which prevents log stampeding in environments receiving very high traffic.

Also bumps rubocop to get rid of a warning when running.

## How was it tested?

`rspec` and through performance/regression testing.

----

@zackangelo @pedelman @mattolson @junedkazi @chrisboulton

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ sudo: false
 dist: trusty
 rvm:
   - 2.2
-  - 2.3
   - 2.4
   - 2.5
 before_install: gem install bundler -v 1.16.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 
 ### Pending release
 
+- Log ok/validation responses at DEBUG levels to prevent log stampeding in high-volume environments
+
 ### 2.5.1
 
 - Ensure `timeout` is an int when passed as a client option to a gRPC client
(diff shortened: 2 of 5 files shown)


pedelman:
@splittingred should we bump minor version? Might be confusing for people to update and lose their expected logs from version <= 2.5.1.

splittingred:
@pedelman Bumping versions is done in a separate commit, to prevent clobbering with multiple PRs on master, and allow a single, isolated ref for it.

.travis.yml
```diff
@@ -5,7 +5,6 @@ sudo: false
 dist: trusty
 rvm:
   - 2.2
-  - 2.3
```
junedkazi: what about ruby 2.2 ?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 
 ### Pending release
 
+### 2.5.2
```
junedkazi: Do we need to bump version file as well ? or we have more PR's being worked on as part of this release.

.travis.yml
splittingred: We're leaving 2.2 support in because it's still widely used; Ruby 2.3 has issues with bundler globally at the moment so I've dropped it here to allow this to pass.

CHANGELOG.md
splittingred: Dropped that changelog header for now, that will come in separate commit, as per https://github.com/bigcommerce/gruf/pull/70#issuecomment-464858867

PR #71, opened 2019-02-18
EOL Ruby 2.2, bump rubocop to 0.64, rspec to 3.8

splittingred:
## What? Why?

* EOL support for ruby 2.2
* Bumps rubocop to 0.64
* Bumps rspec to 3.8
* Addresses rubocop warnings
* Update activesupport/concurrent-ruby dependencies to have a min version
* Bump version to 2.6.0.pre; releases/2.5.x branch has been created.

## How was it tested?

rspec, running server locally.

----

@bigcommerce/platform-engineering @bigcommerce/oss-maintainers @pedelman @junedkazi


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 
 ### Pending release
 
+- Drop Ruby 2.2 support
```
junedkazi: You need to update the version here we as well to 2.6.0

splittingred: Nope, it's still a pending release, no version # is set yet. You only do that on tags - hence the "Pending Release" header. That allows us flexibility on semver.

junedkazi: Then why are we bumping the version in the version file https://github.com/bigcommerce/gruf/pull/71/files#diff-d78e8222a51844b601691817db97dd6dR17

splittingred: That's a good question - so the reason is we'll set on 2.6.0 because dropping Ruby 2.0 means we need to bump to _at least_ 2.6.0; however, that doesn't necessarily mean we'll end there (master, for example, might jump to 3.0).

PR #72, opened 2019-03-05
Add feature for `.bind` to bind multiple controllers to a single service

nikushi:
Hi, thank you for the great gem.

## What? Why?

The current master allows that a gRPC service can be bound to only one controller. By this change, now `.bind` can be called through multiple controllers to bind them to a single gRPC service class.

Let me explain our use case. Let's say we have a `ProductService`, which concern is `Product` resource and `Price` resource, then we want to implement `GetProduct()` rpc and `GetPrice()` rpc on that service. The proto file looks like below:

```proto
service ProductService {
  rpc GetProduct (GetProductRequest) returns (GetProductResponse);
  rpc GetPrice (GetPriceRequest) returns (GetPriceResponse);
}
```

In order to implement those rpcs, there are two approach by using gruf. One is implementing all the rpcs in a single controller `ProductsController` since the current(master) gruf can only allow 1-1 binding. The second approach is that implementing with two separated controllers, `ProductsController` for `#get_product` action, and `PricesController` for `#get_price`. 

This feature provides developers the second option, since separating controllers by each resource is more like Rails way.

## How was it tested?

The functionality worked in our company's service in development.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 /.bundle/
+/.byebug_history
+/.idea/
 /.yardoc
 /Gemfile.lock
 /_yardoc/
--- a/lib/gruf/configuration.rb
+++ b/lib/gruf/configuration.rb
@@ -28,6 +28,7 @@ module Gruf
       ssl_crt_file: '',
       ssl_key_file: '',
       controllers_path: '',
+      controller_registry: Gruf::Controllers::Registry.new,
       services: [],
       logger: nil,
       grpc_logger: nil,
--- a/lib/gruf/controllers/base.rb
+++ b/lib/gruf/controllers/base.rb
@@ -62,7 +62,7 @@ module Gruf
       #
       def self.bind(service)
         service_class = service.name.constantize
-        Gruf.services << service_class
+        Gruf.services << service_class unless Gruf.services.include?(service_class)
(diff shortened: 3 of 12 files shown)


splittingred:
Hey - thanks! You'll need to rebase from master to pickup the Ruby 2.2 fix.

This is kind of against the procedure that gRPC core takes - all methods should bind to the single controller that serves the service. Generally the approach you'd want here is to just have multiple services (gruf can serve multiple services on one process+port).

Fragmenting it like this could lead to poor compositional behavior where people have methods scattered across controllers with no explicit mapping between them. I'm a little hesitant to enable that kind of behavior.

That said, let me think on this and discuss with a few others to see if this is something we'd want to support. Thanks for your contribution and taking the time to spec this out!

nikushi:
Thanks for the feedback! In a viewpoint of a developer who find actual methods from a service definition, as you mentioned, my change could increase difficulty.

At first, at the starting point of making this request, I thought that I want to have multiple controllers because having a controller per resource is more Rails way according to resource oriented manner that Rails takes. But gRPC is not REST. I should instead follow the manner of gRPC.

nikushi:
Let me close the issue. I decided to have multiple services in my project. Thanks a lot anyway.

PR #73, opened 2019-03-05
Remove support for CRuby 2.2

nikushi:
CI for 2.2.x failed. [Support of Ruby 2.2 has ended](https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/)


nikushi:
Ah, this is duplicated with https://github.com/bigcommerce/gruf/pull/71

PR #80, opened 2019-05-13
Add hook support for executing code paths before a server is started and after a server stops

splittingred:
## What? Why?

Adds hook support for executing code paths before a server is started, and after a server stops, as well as a generic hook registry and base class. This allows for execution of code outside of the normal request chain, which can be used to provide server instrumentation, process monitoring, and more.

```ruby
class MyHook < Gruf::Hooks::Base
  def before_server_start(server:)
    # do my thing before the server starts  
  end
  
  def after_server_stop(server:)
    # do my thing after the server stops
  end
end

# Then in an initializer:

Gruf.configure do |c|
  c.hooks.use(MyHook, option_foo: 'value 123')
end
```

## How was it tested?

`spec/demo_server` has a successful hook implementation. Also ran against a live server and tested proper startup/shutdown, and ensured memory was released properly.

----

@bigcommerce/oss-maintainers @bigcommerce/ruby @bigcommerce/platform-engineering

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,6 +382,46 @@ By default, the ActiveRecord Connection Reset interceptor and Output Metadata Ti
 are loaded into gruf unless explicitly told not to via the `use_default_interceptors` configuration
 parameter.
 
+## Hooks
+
+Hooks, unlike interceptors, are executed outside of the request chain, such as when a server starts
+or stops. They run in FIFO order sequentially and do not wrap one another. They can be used to provide
+custom boot sequences, external instrumentation support, or shutdown alerting. 
+
+You can create a hook by extending the `Gruf::Hooks::Base` class and defining the methods
+on the hook you wish to implement:
+
+```ruby
+class MyHook < Gruf::Hooks::Base
+  def before_server_start(server:)
+    # do my thing before the server starts  
+  end
+  
+  def after_server_stop(server:)
+    # do my thing after the server stops
+  end
+end
+
+# Then in an initializer:
+
+Gruf.configure do |c|
+  c.hooks.use(MyHook, option_foo: 'value 123')
+end
+```
+
+Exceptions raised in hooks will halt the execution chain and bubble up the stack appropriately.
+
+### Available Hook Insertion Points
+
+Current hook insertion points are:
+
+* `before_server_start` - Right before the gRPC server starts
+* `after_server_stop` - Right after the gRPC server is shutdown
+
+Note that exceptions raised in `before_server_start` will halt the execution chain for the remaining 
+`before_server_start` hooks, but will still execute the `after_server_stop` hooks as expected. Exceptions raised
+in `after_server_stop` will prevent further `after_server_stop` hooks from running.
+
 ## Instrumentation
 
 gruf comes out of the box with a couple of instrumentation interceptors packed in:
--- a/lib/gruf/hooks/registry.rb
+++ b/lib/gruf/hooks/registry.rb
@@ -1,0 +1,159 @@
+# frozen_string_literal: true
+
+# Copyright (c) 2017-present, BigCommerce Pty. Ltd. All rights reserved
+#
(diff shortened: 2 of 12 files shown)


spec/gruf/hooks/executor_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+# Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+# WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+# OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+#
+require 'spec_helper'
+
+describe Gruf::Hooks::Executor do
```
pedelman: Might be good to include test case for a hook that raises an exception in both before / after stage.

lib/gruf/hooks/registry.rb
```diff
@@ -0,0 +1,159 @@
...
+        end
+      end
+
+      private
+
+      ##
+      # Handle mutations to the hook registry in a thread-safe manner
+      #
+      def hooks_mutex(&block)
+        @hooks_mutex ||= begin
```
pedelman: I always was told `||=` is not thread safe, is that no longer the case? https://stackoverflow.com/questions/9853086/is-in-ruby-thread-safe

splittingred: Given that this just initializes the mutex if it's not initialized, which will happen before any threading takes place in the server, we should be fine here. It's also worth it to note this is assignment to a private variable in a private method, so unless someone is doing "something very bad" with `send`, it's not likely that will affect us here. (The interceptor registry uses the same exact approach, fwiw.)

spec/gruf/hooks/executor_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    context 'if the second hook raises an exception' do
+      before do
+        allow(hook2).to receive(:send).with(cmd, arguments).and_raise(StandardError)
+      end
+
+      it 'the third hook should not execute, but the first should' do
+        expect(hook1).to receive(:send).with(cmd, arguments).once
+        expect(hook3).to_not receive(:send).with(cmd, arguments)
+        subject rescue nil
```
pedelman: Better to do `expect { subject }.to_raise exception_type`? Also should we log a message?

README.md
```diff
@@ -382,6 +382,36 @@ By default, the ActiveRecord Connection Reset interceptor and Output Metadata Ti
...
+end
+
+# Then in an initializer:
+
+Gruf.configure do |c|
+  c.hooks.use(MyHook, option_foo: 'value 123')
+end
+```
+
+The hook will execute. Exceptions raised in hooks will halt the execution chain and bubble up
```
pedelman: The first sentence seems off. Also, is it worth explicitly mentioning that exceptions raised in before_server_start will still allow for after_server_stop to run, and that after hooks may be run even if the before step didnt.

lib/gruf/hooks/registry.rb
```diff
@@ -0,0 +1,159 @@
...
+          @registry.insert(
+            pos,
+            klass: hook_class,
+            options: options
+          )
+        end
+      end
+
+      ##
+      # Insert an hook after another specified hook
```
pedelman: an hook -> a hook, seen a few instances of this in the PR

splittingred:
@pedelman ♻️ 🙇

PR #96, opened 2020-03-11
Fix issue where CLI --suppress-default-interceptors was ignored

splittingred:
## What? Why?

Due to the way the configuration elements are loaded in Gruf, the CLI executor is not respecting --suppress-default-interceptors. This fixes that issue, and adds tests for it (and all other CLI args).

## How was it tested?

`bin/gruf --suppress-default-interceptors`, prying in to ensure success, and also through rspec tests.

----

@bigcommerce/oss-maintainers @bigcommerce/ruby @bigcommerce/platform-engineering @alikinir

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 
 ### Pending release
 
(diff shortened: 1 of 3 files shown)


alikinir:
Amazing ! 
Thanks a lot for immediate response !

PR #100, opened 2020-03-23
Use zeitwerk loader

namtx:
## What? Why?
In this PR, I just added https://github.com/fxn/zeitwerk as gem runtime_dependency as code loader and removed redundant `require_relative`

## How was it tested?

All specs should be  💚


splittingred:
Hey @namtx - thanks for this! I'd like to test this in a few services, mainly to ensure this doesn't break grpc class loading (as their file name structure for the generated classes is non-standard), and ensure there's no threading issues. I'll get to this this week. Thanks again!

UPGRADING.md
```diff
@@ -84,7 +84,7 @@ Interceptors behave very similarly to outer around hooks, with the following cha
 An base interceptor looks like this:
 
 ```ruby
-class MyInterceptor < ::Gruf::Interceptors::ServerInterceptor
```
splittingred, line 87: Is there a reason for all of these removals? Generally I prefer to be explicit with inheritance so as to prevent namespace collisions with external gems.

lib/gruf/integrations/rails/railtie.rb
```diff
@@ -3,7 +3,7 @@
 module Gruf
   module Integrations
     module Rails
-      class Railtie < ::Rails::Railtie
+      class Railtie < Rails::Railtie
```
splittingred, line 6: This actually causes a fatal error:

> NameError: uninitialized constant Gruf::Integrations::Rails::Railtie

I'd prefer we keep the `::` prefixes across the system if possible, thanks.

namtx: Yah, I understand, I brought all global scope resolution back.

UPGRADING.md
namtx: Firstly, I think the code will look better if we can remove all global scope resolution `::` 
But I'm not 100% sure about conflicting. So I reverted all of them.

splittingred:
On a `bundle install` -

```
Bundler could not find compatible versions for gem "zeitwerk":
  In snapshot (Gemfile.lock):
    zeitwerk (= 2.2.2)

  In Gemfile:
    factory_bot was resolved to 5.1.1, which depends on
      activesupport (>= 4.2.0) was resolved to 6.0.2, which depends on
        zeitwerk (~> 2.2)

    gruf was resolved to 2.8.1, which depends on
      zeitwerk (~> 2.3.0)
```

Looks like you probably need to relax this to `>= 2.2`

namtx:
@splittingred Yah, I kept it at `~> 2.2.2`
https://rubygems.org/gems/zeitwerk/versions/2.2.2

splittingred:
Hi @namtx - this is still not working. You can test by starting a demo server with `./spec/demo_server` and then running commands like:

* `bundle exec rake gruf:demo:get_thing`
* `bundle exec rake gruf:demo:get_things`
* `bundle exec rake gruf:demo:create_things`
* `bundle exec rake gruf:demo:create_things_in_stream`

These should all be successful - they are currently failing with: ```I, [2020-06-15T14:09:29.407603 #40410]  INFO -- : Got response from server in client interceptor rpc.thing_service.create_things of type client_streamer: 1.58ms
rake aborted!
NoMethodError: undefined method `to_h' for #<GRPC::Unavailable: 14:Connection reset by peer>
Did you mean?  to_s```

splittingred:
Closing as this work seems to have stalled - feel free to reopen once rebased and made up to date. Thanks!

PR #104, opened 2020-04-21
ResourceExhausted is missing

JuanyiFeng:
## What? Why?
According to official document https://github.com/grpc/grpc-go/blob/master/codes/codes.go, ResourceExhausted is one of the official gRPC error.

But ResourceExhausted error type is missing in client/errors.rb. Adding it in.

## How was it tested?
Include the new changes with a demo server, nothing should be affected.

### DIFF
--- a/lib/gruf/client/error.rb
+++ b/lib/gruf/client/error.rb
@@ -50,7 +50,7 @@ module Gruf
       class NotFound < Validation; end
       class AlreadyExists < Validation; end
       class OutOfRange < Validation; end
-      
+
       class Cancelled < Error; end
       class DataLoss < Error; end
       class DeadlineExceeded < Error; end
@@ -62,7 +62,7 @@ module Gruf
       class Unavailable < Error; end
       class Unimplemented < Error; end
       class Unknown < Error; end
-      
+
     end
   end
 end
--- a/lib/gruf/client/error.rb
+++ b/lib/gruf/client/error.rb
@@ -50,17 +50,19 @@ module Gruf
       class NotFound < Validation; end
       class AlreadyExists < Validation; end
       class OutOfRange < Validation; end
-      class ResourceExhausted  < Validation; end
+      
       class Cancelled < Error; end
       class DataLoss < Error; end
       class DeadlineExceeded < Error; end
       class FailedPrecondition < Error; end
       class Internal < Error; end
       class PermissionDenied < Error; end
+      class ResourceExhausted < Error; end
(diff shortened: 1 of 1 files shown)


lib/gruf/client/error.rb
```diff
@@ -50,7 +50,7 @@ class InvalidArgument < Validation; end
       class NotFound < Validation; end
       class AlreadyExists < Validation; end
       class OutOfRange < Validation; end
-
+      class ResourceExhausted  < Validation; end
```
splittingred, line 53: `ResourceValidation` isn't really a validation-type error - it's meant to be when the server runs out of workers in its thread pool. Can you have this extend `Error` instead? Also, you have an extra space before the `<`.

JuanyiFeng: Make sense, thank you, making the changes.

lib/gruf/client/error.rb
```diff
@@ -50,17 +50,19 @@ class InvalidArgument < Validation; end
...
       class DeadlineExceeded < Error; end
       class FailedPrecondition < Error; end
       class Internal < Error; end
       class PermissionDenied < Error; end
+      class ResourceExhausted < Error; end
       class Unauthenticated < Error; end
       class Unavailable < Error; end
       class Unimplemented < Error; end
       class Unknown < Error; end
+
```
splittingred, line 65: Can you remove this extra space here?

lib/gruf/client/error.rb
```diff
@@ -50,17 +50,19 @@ class InvalidArgument < Validation; end
       class NotFound < Validation; end
       class AlreadyExists < Validation; end
       class OutOfRange < Validation; end
-
+
```
splittingred, line 53: Can you remove the empty tab space here?

JuanyiFeng: sorry, just realize how unreliable is the web editor of git. Doing it in my IDE.

lib/gruf/client/error.rb
```diff
@@ -57,10 +57,12 @@ class DeadlineExceeded < Error; end
       class FailedPrecondition < Error; end
       class Internal < Error; end
       class PermissionDenied < Error; end
+      class ResourceExhausted < Error; end
       class Unauthenticated < Error; end
       class Unavailable < Error; end
       class Unimplemented < Error; end
       class Unknown < Error; end
+
```
splittingred, line 65: Still an extra space here. Mind cleaning it up?

Issue #108, opened 2020-06-03
fail! raises "NameError: uninitialized constant ActiveSupport::JSON" if active_support/core_ext is required in application


AkihikoITOH:
## Please describe the issue

If an application that uses gruf requires `active_support/core_ext`, the gruf server returns `{ "error": "2 UNKNOWN: NameError: uninitialized constant ActiveSupport::JSON" }` upon `fail! :bad_request`, implying `ActiveSupport::JSON` is somehow unloaded.

With `require 'active_support/core_ext'` commented out, it works just as expected and returns `{ "error": "3 INVALID_ARGUMENT: " }`. Also, requiring `active_support/all` or `active_support/json` won't raise `NameError` and works as expected.

## How to Reproduce

Here are the files that reproduce the behaviour.

#### hello.proto
```proto
syntax = "proto3";

package protos.hello;

service Hello {
  rpc SayHello(HelloInput) returns (HelloOutput) {}
}

message HelloInput {}
message HelloOutput {}
```

#### app.rb
```ruby
require 'gruf'
require 'google/protobuf'

# Returns `{ "error": "2 UNKNOWN: NameError: uninitialized constant ActiveSupport::JSON" }`
require 'active_support/core_ext'

# Fails as expected (`{ "error": "3 INVALID_ARGUMENT: " }`)
# require 'active_support/all'

# Fails as expected (`{ "error": "3 INVALID_ARGUMENT: " }`)
# require 'active_support/json'

Google::Protobuf::DescriptorPool.generated_pool.build do
  add_file('hello.proto', syntax: :proto3) do
    add_message('protos.hello.HelloInput') {}
    add_message 'protos.hello.HelloOutput' do
      optional :message, :string, 1
    end
  end
end

HelloInput = ::Google::Protobuf::DescriptorPool.generated_pool.lookup('protos.hello.HelloInput').msgclass
HelloOutput = ::Google::Protobuf::DescriptorPool.generated_pool.lookup('protos.hello.HelloOutput').msgclass

class MyService
  include GRPC::GenericService

  self.marshal_class_method = :encode
  self.unmarshal_class_method = :decode
  self.service_name = 'protos.hello.Hello'

  rpc :SayHello, HelloInput, HelloOutput
end

class MyController < ::Gruf::Controllers::Base
  bind MyService

  def say_hello
    fail! :bad_request
  end
end

server = Gruf::Server.new(Gruf.server_options)
Gruf.services.each { |s| server.add_service(s) }
[...]

PR #109, opened 2020-06-03
Require active_support/json

AkihikoITOH:
This is a potential fix for #108 .

## What? Why?

gruf server would return an unexpected error `NameError: uninitialized constant ActiveSupport::JSON` if `active_support/core_ext` was required in the application.

Since `ActiveSupport::JSON` is what gruf depends on, I believe it should be required by gruf itself.

## How was it tested?

I installed a local build of this version instead of the original gem and manually tested the behavior with the files that can be found in #108.


splittingred:
This should be addressed in https://github.com/bigcommerce/gruf/pull/110 - closing this PR for now. Thanks!

PR #124, opened 2021-03-02
Add support for installing Gruf under Ruby 3.0

ledbettj:
## What? Why?

[gRPC 1.35](https://github.com/grpc/grpc/releases/tag/v1.35.0) and [Protobuf 3.15](https://github.com/protocolbuffers/protobuf/releases/tag/v3.15.0) (and newer versions of both) now support Ruby 3.0.  That means Gruf should now, hopefully, also be usable under Ruby 3.0.

The gemspec currently pins us to `~> 2.4`, which is equivalent to `>= 2.4, < 3.0` -- this PR does the following:

* loosen that requirement to `>= 2.4, < 3.1` to allow installing Gruf under Ruby 3.0
* ~~update the bundler development dependency.  This is maybe unrelated, except that bundler 2.0 was released January 2019-ish and I didn't want to have to install an older version locally to test this change :)~~ this broke CircleCI so I reverted it.

## How was it tested?

* The test suite passes.
* Smoke testing on ruby 3 locally with some simple gRPC services works as expected.

I can't test this change in production quite yet as I still have a few other dependencies that need to be updated for 3.x, so I can't claim that I know it works perfectly just yet.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,6 +37,12 @@ executors:
       ruby-version:
         type: string
         default: "2.7.1"
+  ruby_3_0:
+    <<: *ruby_env
+    parameters:
+      ruby-version:
+        type: string
+        default: "3.0.0"
 
 commands:
   pre-setup:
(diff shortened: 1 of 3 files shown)


splittingred:
@ledbettj Dangit, you beat me to it. Have been watching the protobuf releases for this. 😄  Thanks for jumping in!

Can you add a note to the changelog with something like, "* Adds support for Ruby 3.0"? We'll probably do a minor release for this changeset.

ledbettj:
> @ledbettj Dangit, you beat me to it. Have been watching the protobuf releases for this. smile Thanks for jumping in!
> 
> Can you add a note to the changelog with something like, "* Adds support for Ruby 3.0"? We'll probably do a minor release for this changeset.

Done!  Thanks for the quick feedback.  Feel free to squash/push anything else here.

splittingred (approved):

splittingred:
Yay! Thanks for contributing @ledbettj !

PR #127, opened 2021-03-15
Allow for float/TimeSpec timeout values on clients

splittingred:
## What? Why?

The current gruf implementation always casts `timeout` to an integer when passed. This prevents setting it to a float value, or to a TimeSpec (such as infinite) value, which is supported by gRPC core. This fixes that functionality and allows floats or TimeSpec objects to be passed, while still maintaining the gruf functionality to allow string timeouts to be set (and auto-cast to floats).

## How was it tested?

rspec + manually.

----

@bigcommerce/ruby @bigcommerce/oss-maintainers @alikinir

### DIFF
--- a/lib/gruf/client.rb
+++ b/lib/gruf/client.rb
@@ -62,7 +62,7 @@ module Gruf
       @opts[:hostname] = @opts.fetch(:hostname, Gruf.default_client_host)
       @opts[:channel_credentials] = @opts.fetch(:channel_credentials, Gruf.default_channel_credentials)
       @error_factory = Gruf::Client::ErrorFactory.new
-      client_options[:timeout] = client_options[:timeout].to_i if client_options.key?(:timeout)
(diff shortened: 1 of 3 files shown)


lib/gruf/client.rb
```diff
@@ -216,5 +216,25 @@ def error_deserializer_class
...
+    def parse_timeout(timeout)
+      if timeout.nil?
+        GRPC::Core::TimeConsts::ZERO
+      elsif timeout.is_a?(GRPC::Core::TimeSpec)
+        timeout
+      elsif timeout.is_a?(Numeric)
+        timeout
+      else
+        timeout.to_f
+      end
```
j05h, line 237: What if the timeout does not respond to `to_f`?

splittingred: Good call, probably need to throw an ArgumentError, rather than let the standard exception be raised

j05h (approved):

PR #131, opened 2021-07-29
Remove activesupport as runtime dependency

ianks:
This patch removed `activesupport` as a runtime dependency to make `gruf` easier to adopt for companies that do not use it

❤️


splittingred:
Hi @ianks - gruf currently depends on ActiveSupport for a few things, such as `.constantize`, `.underscore`, and `delegate`, amongst other things. In order to remove this dependency, you'd have to build that functionality into gruf (not something we really want to do at this time).

PR #140, opened 2021-12-29
Relax required ruby version

such:
## What? Why?

Ruby 3.1 was released.
I used no max version, the same way Rails does (amongst others).

## How was it tested?

Ran rspec with ruby 3.1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,12 @@ executors:
       ruby-version:
         type: string
         default: "3.0"
+  ruby_3_1:
+    <<: *ruby_env
+    parameters:
+      ruby-version:
+        type: string
+        default: "3.1"
 
 commands:
   pre-setup:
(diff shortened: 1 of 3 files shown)


splittingred:
Hey @such - thanks! While I can test and run this fine locally, we require CircleCI tests for each ruby version (see .circleci/config.yml), and this would also need to update the README and CHANGELOG appropriately to adjust the supported versions.

I believe there's an outstanding issue with Ruby 3.1 and bundler and the core `grpc` gem that causes OOMs on install in CircleCI, due to CCI using bundler 2.3.3 (which had a packaging issue that resulted in the runtime being broken there). We may have to wait until CCI updates their ruby 3.1 image to bundler 2.3.4 before proceeding. They're typically fairly fast so I'd expect that within a week or so.

such:
Hey @splittingred.
Thank you for the feedback! I'll make those changes!

In the meantime, I'm also having issues with my ruby 3.1 docker image... Can you point me to a thread that talks about this? Do you know if the issue is specific to `grpc`?

splittingred:
@such I believe the CCI issue is related grpc not having pre-built binaries for Ruby 3.1, causing it to have to compile at install time (which makes the gemset _huge_ for one).

See https://github.com/grpc/grpc/issues/28627

Also, `google-protobuf` does not yet support Ruby 3.1:

https://github.com/protocolbuffers/protobuf/issues/9364

splittingred:
Hello @such ! I've opened #150 to get this to work; though it's noted that https://github.com/grpc/grpc/issues/28627 is still an issue, so initial uncached installations of the `grpc` gem will take a _long_ time.

However, this passes tests and end-to-end suites, so we're good to allow it in gruf now.

splittingred:
This has been merged in #150 - closing this. Thanks!

PR #141, opened 2021-12-29
Fixed interceptor order to be FIFO as documented, instead of FILO

### LINKED ISSUE #139: Wrong interceptor execution order
## Please describe the issue

Your yard documentation for the `Gruf::Interceptors::Context#intercept!` method says that interceptors are used in FIFO order. From your code, it looks like FILO order is used because you use the `pop` method instead of `shift`.

## How to Reproduce

Create two interceptors:

```ruby
class FirstInterceptor < Gruf::Interceptors::ServerInterceptor
  def call
    p 1
    yield
  end
end

class SecondInterceptor < Gruf::Interceptors::ServerInterceptor
  def call
    p 2
    yield
  end
end
```
Then use them, expecting them to be executed in FIFO order

```ruby
Gruf.configure do |config|
  config.interceptors.clear
  config.interceptors.use(FirstInterceptor)
  config.interceptors.use(SecondInterceptor)
end
```

After we make the request we will see `2 1` instead of `1 2`.

## What _should_ happen?

We should see `1 2`

* gruf (2.11.0)
* grpc (1.41.0)
* grpc-tools (1.41.0)
* MRI 3.0.2
* Linux

splittingred:
## What? Why?

Interceptors were executing in FILO (first in, last out) order, rather than FIFO (first in, first out) order, which was contrary to documented behavior. This fixes that, and adds a note to the UPGRADING guide.

Because this is a breaking change, we can't fix this in a patch release, so I've bumped this to 2.12.x. Since this fix aligns the code with desired _documented_ behavior, we don't need a major release.

This fixes #139 

## How was it tested?

New specs and a manual test.


UPGRADING.md
```diff
@@ -1,5 +1,12 @@
 This document guides on how to upgrade between significant versions of Gruf.
 
+## Upgrading to 2.12
+
+Gruf 2.12.x fixes a bug in prior versions where interceptors were executed in FILO order, rather than FIFO order, which
+was the documented behavior. If uou are using interceptors, and you believed the order of their execution was in FILO
```
j05h, line 6:
```suggestion
was the documented behavior. If you are using interceptors, and you believed the order of their execution was in FILO
```

j05h (approved):
Spelling error, otherwise 👍

PR #148, opened 2022-02-16
Fix issue with race condition in server starts where servers may fail to bind connections and never reach serving state

### LINKED ISSUE #147: Gruf 2.13 seems to fail to allow connections
## Please describe the issue

After updating to Gruf 2.13, I can no longer make connections from my gRPC clients to my gRPC server. With Gruf 2.12, everything works fine - but with 2.13, it seems like connections are just never serviced. I see the following error reported by my client:
```
[9a625990-be2e-41e0-be7f-638fed11094e] [::1] Gruf::Client::Errors::Unavailable (14:failed to connect to all addresses. debug_error_string:{"created":"@1645049640.389778000","description":"Failed to pick subchannel","file":"src/core/ext/filters/client_channel/client_channel.cc","file_line":3135,"referenced_errors":[{"created":"@1645049640.389777000","description":"failed to connect to all addresses","file":"src/core/lib/transport/error_utils.cc","file_line":163,"grpc_status":14}]}):
```

I don't see any logging in the server process, not even any logs about starting up/etc. (That appears unchanged in both 2.12 and 2.13, but it makes it harder for me to diagnose where the problem might be)

The truly odd part is that this *feels* like an error in the gRPC library itself somehow, given the error message. But I haven't changed gRPC versions, just gruf versions - and when I try upgrading gRPC as well, to 1.43.1, I seem to see the same behavior. I saw notes in previous Gruf releases about gRPC incompatibility, and figured maybe it was related, but thus far, no dice.

## How to Reproduce

I don't know *quite* how to describe this for reproduction in another application - but in my application, it's very simple:
* After spinning up the gRPC server with `bundle exec gruf`, attempt to make any gRPC request via `Gruf::Client.new(<options>).call`
* Observe that the connection never completes and eventually times out with the error above.

## What _should_ happen?

The server should accept connections from the calling client.

* gRPC version - 1.41.0
* Ruby version - 3.0.2
* OS - OSX 11.6.2, M1 MBP

splittingred:
## What? Why?

Fixes issues introduced in ef139b7567f0387cf784ddc7e11f81d62eaea41f, fixes #147.

There seems to be a race condition when starting a `GRPC::RpcServer` in a thread, and attempting to use `server.wait_till_running` in the parent thread - `wait_till_running` never proceeds, even after the server is started. Us calling `server_thread.join` _after_ the `wait_till_running` call compounds this (likely the issue that the Thread never finishes executing and binding the connections before we proceed to a serving state).

This means that 2.13.0 is, in certain cases, starting servers that fail to establish serving connections properly.

This PR removes the wait_till_running call, and just joins on the Thread, allowing gRPC to handle the signals internally. While this means we will be updating the process title to `serving` potentially nanoseconds _prior_ to it actually being running, that seems like a minuscule, acceptable cost given that gRPC does not accept a post-execution block to its run method.

## How was it tested?

`script/e2e`, rspec functional tests, as well as manual testing through `spec/demo_server` and running RPC calls against it.

### DIFF
--- a/lib/gruf/server.rb
+++ b/lib/gruf/server.rb
@@ -90,7 +90,6 @@ module Gruf
         logger.info { "Starting gruf server at #{@hostname}..." }
         server.run_till_terminated_or_interrupted(KILL_SIGNALS)
       end
-      server.wait_till_running
       @started = true
       update_proc_title(:serving)
       server_thread.join
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 
 ### Pending release
 
+### 2.13.1
+
+- Fix issue with race condition in server starts where servers may fail to bind connections and never reach  
+  serving state (fixes #147)
+ 
 ### 2.13.0
 
 - Remove server mutex handling in deference to core grpc signal handling
--- a/spec/gruf/client_spec.rb
+++ b/spec/gruf/client_spec.rb
@@ -145,7 +145,10 @@ describe Gruf::Client do
       let(:method_name) { :GetThing }
 
       it 'calls the appropriate method with the right signature' do
-        expect(client).to receive(:get_thing).with(req, return_op: true, metadata: metadata).and_return(op)
+        expect(client).to receive(:get_thing).with(
+          an_instance_of(::Rpc::GetThingRequest),
+          { return_op: true, metadata: metadata }
+        ).and_return(op)
         expect(subject).to be_truthy
       end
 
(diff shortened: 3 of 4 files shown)


lib/gruf/server.rb
```diff
@@ -90,9 +90,8 @@ def start!
         logger.info { "Starting gruf server at #{@hostname}..." }
         server.run_till_terminated_or_interrupted(KILL_SIGNALS)
       end
-      server.wait_till_running
```
xyzjace, line 93: What happens if the server fails to bind on a socket or similar? Will it raise an exception and prevent us having a hung server?

splittingred: The thread should propagate the error up, as per GRPC, which maintains a mutex and synchronized state on the underlying channel and server. (They're essentially now doing in gRPC core what gruf was doing prior to 2.13.0.) The `wait_till_running` is unnecessary - we were only using it to try and maintain 100% accuracy on the process title (see the PR description for more details).

xyzjace (approved):

PR #150, opened 2022-03-18
Add support for Ruby 3.1

splittingred:
## What? Why?

Adds support for Ruby 3.1.

## How was it tested?

rspec, e2e, manual.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ up fast and efficiently at scale. Some of its features include:
   still preserving gRPC BadStatus codes
 * Server and client execution timings in responses
 
-gruf currently has active support for gRPC 1.10.x+. gruf is compatible and tested with Ruby 2.2-3.0.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -17,7 +17,7 @@ up fast and efficiently at scale. Some of its features include:
   still preserving gRPC BadStatus codes
 * Server and client execution timings in responses
 
-gruf currently has active support for gRPC 1.10.x+. gruf is compatible and tested with Ruby 2.2-3.0.
+gruf currently has active support for gRPC 1.10.x+. gruf is compatible and tested with Ruby 2.2-3.1.
```
j05h, line 20: Is `2.2` really accurate anymore if we're not actively testing it against that (or any other versions) previous to `2.6`?

j05h (approved):

README.md
splittingred: That's a good call, we should bump this to 2.6.

PR #154, opened 2022-05-23
Fix rubocop configuration and failing cop

j05h:
## What? Why?

The `rubocop` job in CircleCI was not running, and instead was calling `bundle-audit`  a second time. This PR corrects that and fixes a found `rubocop` error with `Style/FetchEnvVar` in `Gruf::Configuration`.

## How was it tested?

Run `rubocop` and `rspec` locally.

### DIFF
--- a/spec/gruf/configuration_spec.rb
+++ b/spec/gruf/configuration_spec.rb
@@ -39,10 +39,18 @@ describe Gruf::Configuration do
   describe '.environment' do
     subject { obj.send(:environment) }
 
+    let!(:previous_rack_env) { ENV.fetch('RACK_ENV', nil) }
+    let!(:previous_rails_env) { ENV.fetch('RAILS_ENV', nil) }
+
+    after do
+      ENV['RACK_ENV'] = previous_rack_env
+      ENV['RAILS_ENV'] = previous_rails_env
+    end
+
     context 'with ENV RAILS_ENV' do
       before do
-        allow(ENV).to receive(:[]).with('RACK_ENV').and_return nil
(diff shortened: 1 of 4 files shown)


spec/gruf/configuration_spec.rb
```diff
@@ -39,10 +39,18 @@ class TestConfiguration
...
+      ENV['RACK_ENV'] = previous_rack_env
+      ENV['RAILS_ENV'] = previous_rails_env
+    end
+
     context 'with ENV RAILS_ENV' do
       before do
-        allow(ENV).to receive(:[]).with('RACK_ENV').and_return nil
-        allow(ENV).to receive(:[]).with('RAILS_ENV').and_return 'production'
+        ENV['RACK_ENV'] = nil
+        ENV['RAILS_ENV'] = 'production'
```
splittingred, line 53: Setting ENV in tests will have other issues - we should instead use `stub_const` for this. So something like:

```ruby

  describe '#environment' do
    subject { obj.send(:environment) }

    let(:env_vars) { {} }

    before do
      stub_const('ENV', env_vars)
    end

    context 'ENV RAILS_ENV' do
      let(:env_vars) { { 'RACK_ENV' => nil, 'RAILS_ENV' => 'production' } }

      it 'returns the proper environment' do
        expect(subject).to eq 'production'
      end
    end

    context 'ENV RACK_ENV' do
      let(:env_vars) { { 'RACK_ENV' => 'production', 'RAILS_ENV' => nil } }

      it 'returns the proper environment' do
        expect(subject).to eq 'production'
      end
    end
  end
```

splittingred (changes requested):

splittingred (approved):

PR #161, opened 2022-07-10
Add Rails Integration

chizhanov:
## What? Why?
Hi. I added integration with Rails(since 6.1) [based](https://github.com/anycable/anycable-rails/blob/master/lib/anycable/rails/railtie.rb#L63) on AnyCable. 
Now there is no need to call `bundle exec gruf`.
I had to change the structure of the startup logic a lot, but I hope I didn't break anything.

## How was it tested?
For testing, I used a demo project and tests.

### DIFF
--- a/lib/gruf/server.rb
+++ b/lib/gruf/server.rb
@@ -23,8 +23,6 @@ module Gruf
   class Server
     class ServerAlreadyStartedError < StandardError; end
 
-    KILL_SIGNALS = %w[INT TERM QUIT].freeze
-
     include Gruf::Loggable
 
     # @!attribute [r] port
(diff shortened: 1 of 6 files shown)


splittingred:
@IlyaChizhanov Thanks for the suggestion! I'm a bit concerned with doing this, as this changes the core operating model of gruf, and runs it in a way separate from its original intent. This would be a breaking (major) change, and I'd rather keep gruf separate from the runtime model of rails.

That said, this sounds like an excellent candidate for a gruf-rails gem, which could offer an alternative operating model, one more embedded like ActionCable or other systems. I'd take that path instead.

lib/gruf/server.rb
```diff
@@ -89,31 +86,55 @@ def server
     # Start the gRPC server
     #
     # :nocov:
-    def start!
-      update_proc_title(:starting)
+    def start
```
IlyaChizhanov, line 89: But in any case, I need changes in this part so as not to cause blocking of the main thread.

PR #164, opened 2022-08-03
Add `context` hash attribute to `Gruf::Controllers::Request` to allow interceptors to pass information down to a gruf controller

### LINKED ISSUE #163: Request context for interceptors to communicate information back to controllers
I'm writing interceptor code to parse an access token, and communicate back the parsed token object to the gruf controller, in some capacity, that is only visible in the scope of a single request. Ideally, I would like to set the data in an interceptor, something like:

```
request.context['parsed_token'] = parse_token(token_str)
```

And then access this object in a controller, like:

```
ctx = request.context['parsed_token']
parsed_token.some_value
```

I can accomplish this with `request.active_call.metadata`, but that will (I think) communicate the data back to the client, which is not the intent. It's solely meant as an internal convenience, and to make sure the token isn't parsed more than once. I have to parse it in the interceptor, regardless.

Does gruf provide something like that, or am i thinking of this the wrong way?

Thanks for a great library.

splittingred:
## What? Why?

Adds a `context` attribute to `Gruf::Controllers::Request` that allows Server Interceptors to pass information between each other and down to gruf controllers. Fixes #163.

## How was it tested?

rspec, e2e test, console

### DIFF
--- a/lib/gruf/controllers/request.rb
+++ b/lib/gruf/controllers/request.rb
@@ -36,6 +36,11 @@ module Gruf
       # @!attribute [r] service
       #   @return [Class] The GRPC service class for this request
       attr_reader :service
+      # @!attribute [r] context
+      #   @return [::ActiveSupport::HashWithIndifferentAccess] An arbitrary hash of key/value entries that are
+      #     accessible for interceptors, that can be used to shared information between interceptors and pass down into
+      #     the controller.
+      attr_reader :context
 
       delegate :metadata, to: :active_call
       delegate :messages, :client_streamer?, :server_streamer?, :bidi_streamer?, :request_response?, to: :type
@@ -72,6 +77,7 @@ module Gruf
         @message = message
         @rpc_desc = rpc_desc
         @type = Type.new(rpc_desc)
+        @context = ::ActiveSupport::HashWithIndifferentAccess.new
       end
 
       ##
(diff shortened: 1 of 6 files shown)


lib/gruf/controllers/request.rb
```diff
@@ -36,6 +36,10 @@ class Request
       # @!attribute [r] service
       #   @return [Class] The GRPC service class for this request
       attr_reader :service
+      # @!attribute [r] context
+      #   @return [Class] An arbitrary hash of key/value entries that are accessible for interceptors, that can be
```
billthompson, line 40:
```suggestion
      #   @return [Hash] An arbitrary hash of key/value entries that are accessible for interceptors, that can be
```

billthompson (changes requested):
Overall LGTM. Suggestion for comment update submitted.

I do have a question about this shared context - since interceptors could write keys inconsistently (string and/or symbol), do we want to expose the underlying data structure as we are doing now or provide accessors/readers to enforce consistency?

splittingred:
@billthompson Good point; I think using `::ActiveSupport::HashWithIndifferentAccess` would solve that more efficiently than adding a restricted state object (we already include AS in gruf, so it's trivial to add).

billthompson (approved):
Awesome; LGTM!

PR #165, opened 2022-08-03
Drop Ruby 2.6 support (EOLed)

splittingred:
## What? Why?

Drops support for Ruby 2.6 (EOLed Mar 31 2022).


billthompson (approved):
LGTM.

splittingred:
> I see that the e2e tests run on 3.0 while 2.7 is the default everywhere else. I don't think there's an issue with this but stuck out while reviewing.

The e2e tests (and all test runners) run on each version; the commit was just adjusting the default circle config value to align with the others (3.0).

PR #166, opened 2022-08-03
Allow passing a specific list of services to run via the gruf binstub

### LINKED ISSUE #156: Starting Gruf and bind with specific controllers
Is there a way with Gruf to start a server and bind it to a set of specific controllers within the codebase? I'm currently working on implementing this into a monolith and we would like to have some flexibility regarding what's listened to or not.

```
# example
bundle exec gruf --only=RoleAndPermissionsController
```

Or something of sort, so that the gRPC server doesn't listen to other services when being started.

splittingred:
## What? Why?

Allow passing a specific list of service classes to use when starting a gruf server. Fixes #156.

Can be done via:

```
bundle exec gruf --services NameOfMyRpcServiceClass
```

## How was it tested?

rspec, and:

```
LOG_LEVEL=debug spec/demo_server --services "Rpc::ThingService::Service"
```


spec/gruf/cli/executor_spec.rb
```diff
@@ -94,15 +94,47 @@
       end
     end
 
+    context 'when services are set via the CLI argument' do
+      let(:options_services) { [] }
+      let(:global_services) { [] }
+      let(:cli_services) { [Rpc::ThingService::Service] }
+      let(:args) { %w[--services Rpc::ThingService::Service] }
+
+      it 'adds each specified service to the server' do
```
billthompson, line 103: It would be great if we could populate `options_services` and `global_services` with unique, valid values and ensure that they aren't loaded.

spec/gruf/cli/executor_spec.rb
```diff
@@ -94,15 +94,47 @@
...
+      end
+
+      context 'when the CLI argument has a service that does not exist' do
+        let(:args) { %w[--services FakeServiceThatIsNotReal::AtAll] }
+
+        it 'does not run the server' do
+          expect(server).not_to receive(:start!)
+          expect(hook_executor).not_to receive(:call).with(:before_server_start, server: server)
+          expect(hook_executor).not_to receive(:call).with(:after_server_stop, server: server)
+          expect { subject }.to raise_error(SystemExit)
```
billthompson, line 121: It would be nice if we could assert that the system exited with the human readable error message we expected for the given test (here and on L137).

billthompson (approved):
Overall LGTM. A couple non-blocking nits on the tests.

cwalsh (approved):

PR #177, opened 2022-11-21
Fix Style/GuardClause cop in autoloader

splittingred:
## What? Why?

Just some code style tidying in the autoloader to make Rubocop happy.

## How was it tested?

`bundle exec rubocop -P`


j05h (approved):
Happy cops are silent cops.

Issue #174, opened 2022-11-14
Autoloading in development is not thread-safe


tmtrademarked:
## Please describe the issue

When running a grpc server with more than one thread locally, multiple simultaneous requests can come in at once. In this case, each request will trigger a call to `Gruf.autoloaders.reload`, which attempts to reload the Zeitwerk environment. The result is that while processing requests, some calls will inconsistently fail with Ruby class loading errors.

## How to Reproduce

It was hard for me to reproduce this reliably. Eventually what I did was to create a test script that would invoke my GRPC endpoint, and ran it on a tight loop:
```
for i in {1..10}; do ruby tools/grpc_test_cart.rb 2411031 &; done
```

This results in *some* successful requests, and some failures with errors like the following:
```
.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/grpc-1.50.0/src/ruby/lib/grpc/generic/active_call.rb:29:in `check_status': 2:NameError: uninitialized constant CartServiceRpcController (GRPC::Unknown)

      Object.const_get(camel_cased_word)
            ^^^^^^^^^^. debug_error_string:{UNKNOWN:Error received from peer ipv6:%5B::1%5D:50051 {grpc_message:"NameError: uninitialized constant CartServiceRpcController\\n\\n      Object.const_get(camel_cased_word)\\n            ^^^^^^^^^^", grpc_status:2, created_time:"2022-11-14T15:18:44.571699-05:00"}}
	from /Users/tom.wilson/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/grpc-1.50.0/src/ruby/lib/grpc/generic/active_call.rb:186:in `attach_status_results_and_complete_call'
	from /Users/tom.wilson/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/grpc-1.50.0/src/ruby/lib/grpc/generic/active_call.rb:382:in `request_response'
	from /Users/tom.wilson/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/grpc-1.50.0/src/ruby/lib/grpc/generic/client_stub.rb:180:in `block in request_response'
	from /Users/tom.wilson/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/grpc-1.50.0/src/ruby/lib/grpc/generic/interceptors.rb:170:in `intercept!'
[...]

PR #178, opened 2022-12-28
Introduce read-write lock for code reloading; do code reloading before controller name resolution

leonid-shevtsov:
## What? Why?

- Fixes https://github.com/bigcommerce/gruf/issues/174 (at least, for us)
- Zeitwerk suggests [using a read-write lock for code reloading](https://github.com/fxn/zeitwerk#thread-safety), so that reloading cannot happen during a parallel request. 
- Moved code reloading to happen _before_ the controller is constantized, because otherwise you get a stale class and it doesn't work with reloading included modules.
- Or, rather, now it _wraps_ the entire service method call, first reloads the code if necessary, then locks for read access
- In production mode the write lock will never be acquired, and the read lock will simply let code through.

## How was it tested?

- Added tests, plus errors described in #174 went away on my machine.

### DIFF
--- a/lib/gruf/controllers/autoloader.rb
+++ b/lib/gruf/controllers/autoloader.rb
@@ -15,6 +15,8 @@
 # COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
 # OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #
+require 'concurrent/atomic/read_write_lock'
+
 module Gruf
   module Controllers
     ##
@@ -49,8 +51,16 @@ module Gruf
       def reload
         return unless @reloading_enabled
 
-        reload_mutex do
+        reload_lock.with_write_lock do
           @loader.reload
+        end
+      end
+
+      def with_fresh_controller(controller_name)
+        Gruf::Autoloaders.reload
+        reload_lock.with_read_lock do
+          controller = controller_name.constantize
+          yield controller
         end
       end
 
@@ -77,12 +87,8 @@ module Gruf
       ##
       # Handle thread-safe access to the loader
       #
-      def reload_mutex(&block)
-        @reload_mutex ||= begin
-          require 'monitor'
-          Monitor.new
-        end
-        @reload_mutex.synchronize(&block)
+      def reload_lock
+        @reload_lock ||= Concurrent::ReadWriteLock.new
       end
     end
   end
--- a/lib/gruf/controllers/autoloader.rb
+++ b/lib/gruf/controllers/autoloader.rb
@@ -57,10 +57,11 @@ module Gruf
       end
 
       def with_fresh_controller(controller_name)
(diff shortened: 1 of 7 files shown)


leonid-shevtsov:
Not sure what i should do with the rubocop & codeclimate errors; i could refactor ServiceBinder, but it seems like the issue has been there before my changes. Rubocop issues are old as well.

lib/gruf/controllers/autoloader.rb
```diff
@@ -49,11 +51,19 @@ def initialize(path:, reloading: nil, tag: nil)
...
         return unless @reloading_enabled
 
-        reload_mutex do
+        reload_lock.with_write_lock do
           @loader.reload
         end
       end
 
+      def with_fresh_controller(controller_name)
+        Gruf::Autoloaders.reload
```
splittingred, line 60: IMO, we should probably have this read:

```ruby
def with_fresh_controller(controller_name)
  return yield(controller_name.constantize) unless @reloading_enabled

  ::Gruf::Autoloaders.reload
  reload_lock.with_read_lock do
    yield(controller_name.constantize)
  end
end
```

I realize this does the extra check here, which is done in Gruf::Controllers::Autoloader, but the reason is two-fold:
1) We avoid calling the autoloader entirely here, which saves some CPU clicks
2) We avoid the RW lock entirely, which has a significant performance impact:

I ran a quick script like so:

```ruby
require 'concurrent/atomic/read_write_lock'
rw = Concurrent::ReadWriteLock.new
cycles = 10_000
Benchmark.bm do |x|
  x.report("with_read_lock") do
    cycles.times do
      rw.with_read_lock do
        Math.sqrt(4)
      end
    end
  end
  x.report("w/o_read_lock") do
    cycles.times do
      Math.sqrt(4)
    end
  end
end
```

This produced results:

```
       user     system      total        real
with_read_lock  0.013227   0.000035   0.013262 (  0.013283)
w/o_read_lock  0.000483   0.000001   0.000484 (  0.000484)
```

This is statistically significant  enough for us to want to ensure that execution path does not run - so we should bail early in the method.

splittingred (changes requested):
Thanks for doing this! A few small tweaks to improve performance.

As per the Rubocop error, you can add this to the .rubocop.yml file:

```yaml
Style/RedundantConstantBase:
  Enabled: false
```

We're fine with full path resolution (I prefer it as it enables safer refactoring).

lib/gruf/controllers/autoloader.rb
leonid-shevtsov: Wow you're right - that's an order-of-magnitude difference!

Taking it a bit further - should we also skip `.name.constantize` when there is no reloading? Constantize is not a free operation, either, and if classes don't change, there's no reason to re-resolve the controller on every call.

splittingred: IIRC the `.constantize` there is necessary to ensure that Zeitwerk picks it up from an autoloading perspective. We should keep that there.

leonid-shevtsov:
Updated to bypass locking mechanism

splittingred:
@leonid-shevtsov Thanks! https://github.com/bigcommerce/gruf/pull/181 should fix the rubocop issue.

leonid-shevtsov:
Sorry, got back from the break. All done. Pleased to resolve my own rubocop warnings.

splittingred (approved):

PR #183, opened 2023-02-15
Dynamically determine use_default_interceptors

heka1024:
## What? Why?
- I want to disable default interceptors 
- But `use_default_interceptors`'s default value is `true` so there are always default interceptor after `Configuration#reset` is called.
- To fix that behavior, add `GRUF_USE_DEFAULT_INTERCEPTORS` ENV to dynamically enable/disable default interceptors

## How was it tested?
- From local fork 😄

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 
 ### Pending release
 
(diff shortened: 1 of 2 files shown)


splittingred (approved):
Looks good to me. Can you rebase from main to pickup the CircleCI rubocop fix?

heka1024:
@splittingred Sure!

PR #190, opened 2023-11-14
Fix runtime dependencies

y-yagi:
## What? Why?

It seems that unsued gems are defined in `runtime_dependency`. This PR did the following.

* Remove `thwait` from `runtime_dependency` which is only used in tests.

```bash
$ git grep thwait
Gemfile:gem 'thwait', '>= 0.1'
spec/gruf/synchronized_client_spec.rb:require 'thwait'
```

* Remove `e2mmap` from `runtime_dependency`. This gem was added by f157736,  but it seems isn't used now.

## How was it tested?

I confirmed that the specs are passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 
 ### Pending release
 
+* [#190] Remove unsued `e2mmap` and `thwait` gems from `runtime_dependency`.
+
 ### 2.19.0
 
 * [#197] Add support for Ruby 3.3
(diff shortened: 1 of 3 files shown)


splittingred:
Hi, thanks for this! Can you add a note to CHANGELOG.md describing this change? Thanks!

y-yagi:
@splittingred 
Thanks for your comment! I added a CHANGELOG entry.

splittingred:
Hi @y-yagi - can you rebase from latest?

y-yagi:
@splittingred 
Done!

splittingred (approved):

PR #193, opened 2023-11-28
#191. Allow passing of request object directly

### LINKED ISSUE #191: Allow gruf to receive request objects in `call`
## What? Why?

Currently, `gruf` only supports hashes when sending messages - it won't allow you to actually pass in the request object itself. E.g. in [client_spec.rb](https://github.com/bigcommerce/gruf/blob/main/spec/gruf/client_spec.rb#L136), this doesn't work:

```ruby
    subject { client.call(method_name, req, metadata, opts) }
```

We'd really like to use the request object directly because we can make use of Sorbet generated types when we do this. We also can't just call `to_h` on the object and pass it in, because there are bugs in this method (such as [this one](https://github.com/protocolbuffers/protobuf/issues/6167)).

## Other Notables

It seems like this would be simple enough to implement by just checking to see if the passed in `params` respond to the `to_proto` message and if so, not to call the `request_object` method and just use it as is.

I can have a quick PR with this up if there are no objections.

dorner:
## What? Why?

Fixes #191. This allows passing the request object directly without first converting it to a hash.

## How was it tested?

Unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 ### Pending release
 
 * Add `GRUF_USE_DEFAULT_INTERCEPTORS` ENV to dynamically enable/disable default interceptors
+* Allow passing of request object directly into `call`
 
 ### 2.17.0
 
--- a/lib/gruf/client.rb
+++ b/lib/gruf/client.rb
@@ -81,7 +81,11 @@ module Gruf
     #
     def call(request_method, params = {}, metadata = {}, opts = {}, &block)
       request_method = request_method.to_sym
-      req = streaming_request?(request_method) ? params : request_object(request_method, params)
+      req = if params.respond_to?(:to_proto) || streaming_request?(request_method)
+              params
+            else
+              request_object(request_method, params)
+            end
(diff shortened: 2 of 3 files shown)


splittingred (commented):
Excellent, thanks! Can you add a short note to CHANGELOG.md describing the changes?

dorner:
Added!

splittingred (approved):

dorner:
Awesome, thanks so much! Not sure what the timelines are for releases - when do you think I can expect a version bump? :grin:

splittingred:
@dorner We're about due for a minor release for this lib, so:

https://github.com/bigcommerce/gruf/releases/tag/v2.18.0
https://rubygems.org/gems/gruf/versions/2.18.0

dorner:
YAHOO! Thanks so much!

splittingred:
We don't have specified timelines for this library, mainly because it's mostly feature-complete and doesn't get a lot of changes more than maintenance at this point (gruf does a "small-core" model, which allows for plugins/hooks/interceptors to attach into it, allowing us to keep the core nimble and quick). I typically look at what's pending, and decide from there if we're ready for a minor or patch release accordingly.

PR #194, opened 2023-12-09
add a code reload interceptor for rails

ximus:
Currently only controller code gets reloaded when using rails.

Code reload has worked well for me using this patch. Both controllers and other rails code get reloaded well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Changelog for the gruf gem. This includes internal history before the gem was ma
 ### Pending release
 
 * [#190] Remove unsued `e2mmap` and `thwait` gems from `runtime_dependency`.
+* [#194] Add interceptor to reload Rails app code accross requests
 
 ### 2.19.0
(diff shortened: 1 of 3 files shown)


splittingred:
Hi @ximus - this generally looks good, as `.reloader` has been in Rails since 6.0, so I think we're OK to support it here.

Can you rebase from latest? Also, I think you're hitting this issue regarding CircleCI: https://support.circleci.com/hc/en-us/articles/12504745606171-Why-am-I-seeing-the-Could-not-find-a-usable-config-yml-you-may-have-revoked-the-CircleCI-OAuth-app-message

ximus:
@splittingred rebased

looks like I was able to fix the circleci thing, thanks for the link. 

lmk if you want to deal with the code climate issue and refactor `reset` here.

splittingred (approved):
I think we're fine with the CC thing, it's subjective. Thanks for doing this. :+1:

PR #197, opened 2023-12-26
support ruby 3.3

ximus:
Ruby 3.3 just released.

This has been working fine for me in my app.

* bump dependency in gempsec
* update circleci config to run 3.3
* bump pry, I ran into https://github.com/pry/pry/issues/2271


gruf.gemspec
```diff
@@ -32,7 +32,7 @@ Gem::Specification.new do |spec|
   spec.executables << 'gruf'
   spec.require_paths = ['lib']
 
-  spec.required_ruby_version = '>= 2.7', '< 3.3'
+  spec.required_ruby_version = '>= 2.7', '< 3.4'
```
gi, line 35: Is there a reason why the ruby version has an upper bound other than confirming with testing?

Removing this upper bound would avoid this issue where this library prevents applications from updating ruby.

splittingred: Given gRPC's often incompatibilities with new Ruby versions, we've pinned this library explicitly until tests can be done.

We'll get a working build of this up this week, with 3.3 support.

heka1024: Happy new year, @splittingred ! I wonder if there is any update on this issue. 😄

ximus:
@Jesterovskiy The CI error is `Could not find a usable config.yml`. I'm not sure what I can fix there, I have two other open PRs and they have the same error. Something with how ci is configured in this repo.

I'll let @splittingred weigh in as I'm not a core contributor here.

splittingred:
Just an FYI here: https://github.com/bigcommerce/gruf/pull/199 

Also, I'm not sure what's going on with the CircleCI implementation here - it seems related to forks.

splittingred:
Closing, fixed by #199

PR #198, opened 2023-12-28
Relax ruby requirements

such:
Add ruby 3.3 support

## What? Why?

Ruby 3.3 was released (GA)

## How was it tested?

All tests are green

### DIFF
--- a/gruf.gemspec
+++ b/gruf.gemspec
@@ -32,7 +32,7 @@ Gem::Specification.new do |spec|
   spec.executables << 'gruf'
   spec.require_paths = ['lib']
 
-  spec.required_ruby_version = '>= 2.7', '< 3.3'
+  spec.required_ruby_version = '>= 2.7', '< 3.4'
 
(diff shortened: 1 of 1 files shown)


gruf.gemspec
```diff
@@ -32,7 +32,7 @@ Gem::Specification.new do |spec|
   spec.executables << 'gruf'
   spec.require_paths = ['lib']
 
-  spec.required_ruby_version = '>= 2.7', '< 3.3'
+  spec.required_ruby_version = '>= 2.7', '< 3.4'
```
gi, line 35: Is there a reason why the ruby version has an upper bound other than confirming with testing?

Removing this upper bound would avoid this issue where this library prevents applications from updating ruby.

Same comment here: https://github.com/bigcommerce/gruf/pull/197#discussion_r1441012703

such:
Duplicate of https://github.com/bigcommerce/gruf/pull/197

PR #213, opened 2024-10-31
Support multiple controllers paths

salmanasiddiqui:
## What? Why?

I ran into a problem in a Rails app which was made up of multiple Rails engines. The problem was that two engines were using Gruf and when they were used in same Rails app, the Gruf config for `controllers_path` was getting overridden.

So to fix that problem, I patched Gruf to support multiple paths for controllers by adding a new config field `controllers_paths`. This field defaults to `[controllers_path]`. Gruf uses this `controllers_paths` instead of `controllers_path` everywhere, but `controllers_path` is kept in config for backward compatibility.

Note: I think the config should support something like `add_controllers_paths`, instead of directly setting `controllers_paths`. That will allow multiple modules to add their controllers path without messing with the existing paths. Let me know if you like this idea, I will update the PR

## How was it tested?

All the specs are passing


lib/gruf/autoloaders.rb
```diff
@@ -24,12 +24,12 @@ class << self
...
       ##
-      # Initialize the autoloaders with a given controllers path
+      # Initialize the autoloaders with a given controllers paths
       #
-      # @param [String] controllers_path The path to Gruf Controllers
+      # @param [Array<String>] controllers_paths The path to Gruf Controllers
       #
-      def load!(controllers_path:)
-        controllers(controllers_path: controllers_path)
+      def load!(controllers_paths:)
```
splittingred, line 31: Hi @salmanasiddiqui - while this is a good change to support multiple paths, this unfortunately makes a backwards incompatible change here I'd like to avoid.

Can we add this as a _new_ keyword argument, and deprecate the old one (to be removed in gruf 3.0). We should support both syntaxes in 2.x.

splittingred (changes requested):

lib/gruf/autoloaders.rb
salmanasiddiqui: Hey @splittingred, I have created a new PR with the suggestion you made https://github.com/bigcommerce/gruf/pull/228.

I had to create a new branch in my fork to allow the main branch be synced. Thats why this PR is closed.

PR #216, opened 2024-11-09
Fix integration with Rails when controllers_path is modified

### LINKED ISSUE #215: Gruf.controllers_path config breaks integration with Rails
## Please describe the issue

https://github.com/bigcommerce/gruf/blob/main/lib/gruf/integrations/rails/railtie.rb#L26

The above file adds an intializer which is invoked `before_configuration`. The initializer is supposed to modify the Rails autoloader and ignore `Gruf.controllers_path` from it.

But `Gruf.controllers_path` is supposed to be configured within an intializer. The `before_configuration` hook is too early in the life cycle, and at this stage none of the intializers are executed. So the above file always end up ignoring the default controllers_path from Rails autoloader.

Rails intializer hooks are:
1. before_configuration
2. before_initialize
3. before_eager_load (only executed if eager_load=true)
4. prepare_to (executed on each request in dev environment, executed once on boot in others)
5. after_initialize

Instead of `before_configuration` I think this should depend on eager_load config. And it should be something like

```ruby
if config.eager_load
  before_eager_load { ..."setup the rails autoloader and eager load paths" }
else
  after_initialize { ..."setup the rails autoloader and eager load paths" }
end
```

## How to Reproduce

1. Initialize new rails app
2. Add gruf to it
3. add gruf.rb to config/initializers
```ruby
Gruf.configure do |c|
  c.controllers_path = "app/my_custom_rpc_dir"
end
```
4. create app/my_custom_rpc_dir and add a Grpc service to it
5. run `bundle exec gruf`
6. see error regarding zietwerk conflict

## What _should_ happen?

`bundle exec gruf` should run successfully as the configuration is correct

* gRPC version
* Ruby version
* OS
* Does it occur at certain traffic volumes, payload size, etc

salmanasiddiqui:
## What? Why?

Fixes #215

### DIFF
--- a/lib/gruf/integrations/rails/railtie.rb
+++ b/lib/gruf/integrations/rails/railtie.rb
@@ -23,10 +23,10 @@ module Gruf
       #
       class Railtie < ::Rails::Railtie
         initializer 'gruf.initializer' do |app|
-          config.before_configuration do
+          config.after_initialize do
             # Remove autoloading of the controllers path from Rails' zeitwerk, so that we ensure Gruf's zeitwerk
             # properly manages them itself. This allows us to manage code reloading and logging in Gruf specifically
-            app.config.eager_load_paths -= [::Gruf.controllers_path] if app.config.respond_to?(:eager_load_paths)
+            app.config.eager_load_paths -= [Gruf.controllers_path] if app.config.respond_to?(:eager_load_paths)
             if ::Rails.respond_to?(:autoloaders) # if we're on a late enough version of rails
               ::Rails.autoloaders.each do |autoloader|
                 autoloader.ignore(Gruf.controllers_path)


salmanasiddiqui:
@splittingred 👋🏽 can you give this small fix a look?

lib/gruf/integrations/rails/railtie.rb
```diff
@@ -23,10 +23,10 @@ module Rails
       #
       class Railtie < ::Rails::Railtie
         initializer 'gruf.initializer' do |app|
-          config.before_configuration do
+          config.after_initialize do
```
splittingred, line 26: Rails is pretty finicky with initialization order changes between versions. @salmanasiddiqui - I think we're probably going to need to see some proof of life here that this change doesn't break Rails 7.0, 7.1, 7.2, and 8.0 installations.

splittingred (changes requested):

lib/gruf/integrations/rails/railtie.rb
zzak: I think eager load paths are frozen at the point of `after_initialize` so you won't be able to mutate it

salmanasiddiqui: thanks for reviewing this, Im trying to figure if there is a good way to test this railtie. Any suggestions or relevant resources will be greatly appreciated

@zzak - `after_initialize` is used by gruf-rspec aswell, and I have tested this with Rails 7.1, it updated the autoloader correctly.
https://github.com/bigcommerce/gruf-rspec/blob/main/lib/gruf/rspec/railtie.rb#L25

zzak: Have you tried EAGER_LOAD=true?

salmanasiddiqui: yes, I did

phoffer:
I am working to make a Rails 6.1 app Zeitwerk compatible, and am running into the same issue as #215 with `controllers_path` being configured, and then encountering the Zeitwerk conflict. This patch does fix it for us as well.

PR #222, opened 2025-01-10
Ruby 3.4 support

ximus:
## What? Why?

It's that time of year again.

## How was it tested?

`bundle rake`


ximus:
thought i had checked, but there's already a pr for this, closing

Issue #223, opened 2025-01-29
Lack of Ruby 3.4 support


larouxn:
Ruby 3.4 was released this past December but appears `gruf` is still locked to Ruby `< 3.4`. Thus, opening this issue.

https://github.com/bigcommerce/gruf/blob/db34feeaba84d8b0b0beb9d8bf36b65fdc63c653/gruf.gemspec#L35

- https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-0-released/
- https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-1-released/

splittingred:
Fixed in #221 - was waiting on gRPC 3.4 support and gRPC Ruby 3.4 pre-compiled binaries.

PR #221, opened 2024-12-29
Add support for Ruby 3.4

mmizutani:
## What? Why?

Ruby 3.4 ([3.4.0](https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-0-released/) and [3.4.1](https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-1-released/)) has been released, bringing some improvement in YJIT performance, syntax ergonomics, etc.
This PR ensures that `gruf` gem works with Ruby 3.4.

## How was it tested?

Local testing and the end-to-end suite in this repository were performed to ensure performance and grpc gem compatibility.
Although the CI container image for Ruby 3.4 is not available, at least local unit tests and e2e tests with Ruby 3.4 are passing.

## Prerequisites

- [x] https://github.com/CircleCI-Public/cimg-ruby/pull/190 (superseded by https://github.com/CircleCI-Public/cimg-ruby/pull/194)
- [x] grpc gem (precompiled binary) support for Ruby 3.4
  - [x] https://github.com/grpc/grpc/pull/38601
  - [x] https://github.com/grpc/grpc/pull/38615
  - [x] ~grpc gem 1.70.0 release https://github.com/grpc/grpc/releases/tag/v1.70.0~ grpc gem 1.70.1 release https://github.com/grpc/grpc/releases/tag/v1.70.1

PR #224, opened 2025-01-29
Add support for Ruby 3.4 and drop support for EOL Ruby 3.0

larouxn:
## What? Why?

Adding support for Ruby 3.4, removing support for EOL Ruby 3.0, and adding one RuboCop config entry as with bumping to min Ruby version 3.1 in RuboCop config we've encountered a new default cop that will rename all `&block` to just `&` which I don't think is desirable. Please let me know if we're fine with that change, shown in part below.

Screenshot 2025-01-29 at 19 45 38 (2)

Resolves https://github.com/bigcommerce/gruf/issues/223

## How was it tested?

This was primary just run through the test suite. Beyond that I tested the updated version can be `bundle install`-ed on a Ruby 3.4.1 application. Though haven't been able to actually run such yet as that application has other gems that need to be updated. :grimacing:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,12 +19,6 @@ gem_cache_key: &gem_cache_key
   gem_cache_key: "gem-cache-v2"
 
 executors:
-  ruby_3_0:
-    <<: *ruby_env
-    parameters:
-      ruby-version:
-        type: string
-        default: "3.0"
   ruby_3_1:
     <<: *ruby_env
     parameters:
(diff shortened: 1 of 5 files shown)


larouxn:
e2e test failure looks to be a timeout issue, which can occur when adding a new Ruby as there's now cache and thus a container needs to build all native gems and thus runs out of time. We could temporarily bump the timeout from say 20 minutes to 30 minutes and then reduce once this initial Ruby 3.4 PR is merged.

```console
Installing grpc 1.69.0 with native extensions

Too long with no output (exceeded 10m0s): context deadline exceeded
```

splittingred:
This is a duplicate of #221 - same issue as we are waiting for precompiled grpc gem support from google - @larouxn I'm going to close this, feel free to join the discussion there!

PR #225, opened 2025-02-02
fix request.service_key strip suffix

t0ch1k:
## What? Why?
I don't want service_key to remove all the occurrences of `.service`, only to strip suffix

## How was it tested?
within the existing unit tests + manually


lib/gruf/controllers/request.rb
```diff
@@ -87,7 +87,7 @@ def initialize(method_key:, service:, rpc_desc:, active_call:, message:)
       # @return [String] The mapped service key
       #
       def service_key
-        @service.name.to_s.underscore.tr('/', '.').gsub('.service', '')
+        @service.name.to_s.underscore.tr('/', '.').delete_suffix('.service')
```
splittingred, line 90: Hi @t0ch1k - can we add a unit test that covers your case explicitly? (that it deletes only the suffix, not all instances)

t0ch1k:
@splittingred it seems like merging the latest main into my branch fixed the rubocop issues. Could the PR be merged?

splittingred (approved):

